PR #77, opened 2013-12-30
Added **kwargs to Collection.find method in order to allow MongoEngine t...

castarco:
Added **kwargs to Collection.find method in order to allow MongoEngine to use MongoMock instead of PyMongo.

This change don't break anything, but allows MongoEngine to use MongoMock instead of PyMongo (MongoEngine also requires minor changes that I'm also pull requesting to their repository).

I've added **kwargs as a "general" solution, but the problem I was trying to solve was the need of the parameter slaves_okay .

I'm open to any suggestion to achieve the same result if this change don't seems to you OK.

Thanks for your attention :) .

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -455,7 +455,7 @@ class Collection(object):
             "err": None,
         }
 
-    def count(self):
+    def count(self, with_limit_and_skip = False):
         return len(self._documents)
 
     def drop(self):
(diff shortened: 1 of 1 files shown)


mdomke:
The `slave_okay` parameter of the `find` method has ben removed anyhow so I guess that his PR is obsolete by now. I you think otherwise please re-open it!

PR #96, opened 2014-05-16
Group and Aggregate functionality

jeffreybrowning:
Hey,

I added group and aggregate functions to mongomock.

Not all of the aggregate and group operators are supported, the code throws notimplemented errors when that's the case, but the main operators are supported.

Let me know if you have any questions.

Jeff

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -580,7 +580,8 @@ class Collection(object):
         """)
 
         ret_array = []
-        doc_list_no_id = []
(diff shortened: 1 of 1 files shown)


smalekzadeh:
Hi
I get "'Collection' object is not callable" when trying to test aggregate pipeline. Example

aggresults = mongocdr.aggregate(pipeline)

The code works with the mongodb client but produces the above error when using mogomock in unit tests. can anyone help please?

Thanks

PR #125, opened 2015-01-02
implements $unwind in the aggregation pipeline

zcarter:
also adds some error handling on $group if execjc not installed


vmalloc:
Thanks!

PR #128, opened 2015-01-28
Implemented $setOnInsert

pgy:
Hi,
I wrote tests and implemented the [`$setOnInsert`](http://docs.mongodb.org/manual/reference/operator/update/setOnInsert/) update operation.
It passes all tests with mongodb version 2.6.7 on my machine.


vmalloc:
Thanks!

PR #129, opened 2015-02-12
Add support for _id as dict().

mdprewitt:
Currently, collection throws an 'unhashable' exception in _insert trying
to us data['id]' as a key to the doc dictionary.  Pymongo allows _id to be a
dict().

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -255,11 +255,9 @@ class Collection(object):
                             existing_document['_id'] = _id
                         existing_document.update(self._internalize_dict(document))
                         if existing_document['_id'] != _id:
-                            # id changed, fix index
-                            if type(_id) == dict:
-                                _id = hashdict(_id)
-                            del self._documents[_id]
-                            self.insert(existing_document)
+                            raise OperationFailure(
+                                "The _id field cannot be changed from {0} to {1}".format(
+                                    existing_document['_id'], _id))
                         break
                     else:
                         # can't mix modifiers with non-modifiers in update
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -8,7 +8,7 @@ import warnings
 from sentinels import NOTHING
 from .filtering import filter_applies, iter_key_candidates
 from . import ObjectId, OperationFailure, DuplicateKeyError
-from .helpers import basestring, xrange, print_deprecation_warning
+from .helpers import basestring, xrange, print_deprecation_warning, hashdict
 
 try:
     from collections import OrderedDict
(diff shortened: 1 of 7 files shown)


mongomock/helpers.py
```diff
@@ -48,7 +48,74 @@ def _index_list(key_or_list, direction=None):
...
+    def __key(self):
+        return tuple(sorted(self.items()))
+
+    def __repr__(self):
+        return "{0}({1})".format(
+            self.__class__.__name__,
+            ", ".join("{0}={1}".format(str(i[0]), repr(i[1])) for i in self.__key()))
+
+    def __hash__(self):
+        return hash(self.__key())
```
vmalloc: Note that this doesn't guarantee hashability - all the items (both keys and values) must be hashable for this to work... also consider using frozenset() instead of sorted(), as it is slightly cheaper in run time.

vmalloc:
The travis tests are failing...

mongomock/helpers.py
mdprewitt: I switched to frozenset and also made sure that all values are also converted from dicts as well.  Don't really need to check that keys are hashable as you can't create a dictionary with a non-hashable key in the first place.  For example, python wouldn't let you do this:

```
>>> {{'a': 1}: 1}
  File "<stdin>", line 1, in <module>
TypeError: unhashable type: 'dict'
```

So, only the values are converted from dicts to hashdicts.

vmalloc: dicts are not enough - values can contain lists as well, which are also not hashable

mdprewitt: True.  However, mongo doesn't allow lists for _id: 

http://docs.mongodb.org/manual/core/document/#document-id-field

vmalloc: Not even as values?

mdprewitt: Yes, it does allow values to be lists.  I've updated to handle that.  Also, I updated update/delete to handle dicts in _id.

mongomock/collection.py
```diff
@@ -254,6 +256,8 @@ def update(self, spec, document, upsert = False, manipulate = False,
                         existing_document.update(self._internalize_dict(document))
                         if existing_document['_id'] != _id:
```
mdprewitt: @vmalloc When I was fixing this to handle dict _id's I found that mongo doesn't allow changing `_id` in an update.  It throws this message:

```
The _id field cannot be changed from ...
```

For compatibility, should this throw an exccption instead of delete/insert?

vmalloc: Yes, I think it should

mdprewitt:
I've looked at the failing tests for this and they are all due to a connection reset when trying to download libs.  I didn't see a way to force a rerun.  Unless you have a way, I can force a push to start the rerun.  Otherwise, I think that everything looks good.

vmalloc:
Ok. I will merge it and see about the failures later. Thanks!

PR #134, opened 2015-02-26
Fix for setting limit to 0

albrycht:


vmalloc:
Thanks! Can you please add a test that fails before the change and succeeds after it?

albrycht:
Done.

vmalloc:
Thanks!

PR #135, opened 2015-02-27
Use pip to install dependencies so you'd be able to retry failed installations

thedrow:
This will resolve the network issues we've seen in #129.


vmalloc:
Not sure pip actually retries anything, but it's a welcome change in any case. Thanks!

thedrow:
pip doesn't retry anything but travis will since I specified that the command would be run using travis_retry which retries the same command up to 3 times if it has failed.

PR #136, opened 2015-03-19
Allow subdocument replacement with $set operator

mdomke:
* With {"$set": {"subdocs.2": {...}}} you can replace the third sub-
  document referenced by "subdocs"


vmalloc:
Thanks!

PR #143, opened 2015-06-22
Add $elemMatch return fields operator

magaman384:


vmalloc:
It seems the build is failing on Python 2.6.

mongomock/collection.py
```diff
@@ -354,6 +354,51 @@ def _copy_field(self, obj, container):
         else:
             return copy.copy(obj)
 
+    def _extract_projection_operators(self, fields):
+        """Removes and returns fields with projection operators."""
+        result = {}
+        allowed_projection_operators = {'$elemMatch'}
```
mdomke: I don't know who still uses Python 2.6, but the set syntax with curly braces is not supported there. You'll have to write `set(['$elemMatch'])`.

vmalloc:
The Travis build is failing (I'm guessing on 2.6)...

mdomke:
Yeah. As I pointed out this is because set literals are not available in Python 2.6. If Python 2.6 is still a version that needs to be supported, than you have to use the set constructor explicitly.

mdomke:
@vmalloc In my opinion it would be ok to just drop support for Python 2.6. All reasonably up to date distributions ship with at least Python 2.7.x
@magaman384 Would you mind to rebase your pull request in order to resolve the merge conflicts?

vmalloc:
@mdomke I tend to agree. I see no reason for keeping 2.6 support at this point.

magaman384:
@mdomke No problem, done.

PR #146, opened 2015-07-07
Adaptations to actual pymongo signatures

aviau:
Hello!

Adapting those signatures fixed using MongoMock with MongoEngine for me.

I got it working with:

```
def get_connection(alias):
    return mongomock.Connection()

mongoengine.connection.get_connection = get_connection
mongoengine.connect('db')
```

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -518,10 +518,10 @@ class Collection(object):
                         manipulate, safe, _check_keys = True, **kwargs)
             return to_save.get("_id", None)
 
-    def remove(self, spec_or_id = None, search_filter = None, **kwargs):
+    def remove(self, spec_or_id = None, search_filter = None, **write_concern_kwargs):
         """Remove objects matching spec_or_id from the collection."""
-        if kwargs:
-            write_concern = WriteConcern(**kwargs)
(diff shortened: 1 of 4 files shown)


mongomock/collection.py
```diff
@@ -518,7 +518,7 @@ def save(self, to_save, manipulate = True, safe = False, **kwargs):
                         manipulate, safe, _check_keys = True, **kwargs)
             return to_save.get("_id", None)
 
-    def remove(self, spec_or_id = None, search_filter = None):
+    def remove(self, spec_or_id = None, search_filter = None, **kwargs):
```
vmalloc: I don't like receiving kwargs in such cases. It would make production code passing nonexistent parameters pass tests, but fail in production

aviau: I have now added a WriteConcern object, just like in pymongo.

vmalloc: Can you please rename to write_concern_kwargs to make it clear that's what it's intended for?

aviau: Done

vmalloc:
thanks!

aviau:
Thank you for the quick review :)

PR #157, opened 2015-07-27
Missing values should have lowest priority when sorting.

kilotaras:
As per http://docs.mongodb.org/manual/reference/method/cursor.sort/#ascending-descending-sort

`sentinels.NOTHING` is bigger than everything instead of smaller.


vmalloc:
Thanks!!!

PR #170, opened 2015-08-18
Adds exception raising and test for empty logical operators

julianhille:
I added the same behavior pymongo / mongodb has if the operator sub search is an empty.


mdomke:
Looks good to me.

PR #172, opened 2015-08-20
Allow update with upsert to have an _id field in doc

### LINKED ISSUE #171: update with upsert=True does not accept docs with _id
The following code-snippet would currently raise a `OperationFailure` exception in mongomock but works with pymongo:

```python
from bson import ObjectId
from pymongo import MongoClient

client = MongoClient()
client.test.test.update({'a': 1}, {'_id': ObjectId('55d58d32f4811f67bbaeaf3b'), 'a': 1}, upsert=True)
```

The exception states:

```
OperationFailure: The _id field cannot be changed from 52d669dcad547f059424f783 to 55d50264f4811f64fb918ee8
```

mdomke:
Fixes #171

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -182,7 +182,6 @@ class Collection(object):
         validate_is_mapping('spec', spec)
         validate_is_mapping('document', document)
 
-        found = False
         updated_existing = False
         upserted_id = None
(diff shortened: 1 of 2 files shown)


vmalloc:
Thanks!

PR #174, opened 2015-08-24
Nested pulling and tests added

julianhille:
Pull does not work for nested pulling. Generalized/added logic and adds test to $pull.


julianhille:
theres a mistake with the usage of remove.

PR #177, opened 2015-08-27
Adds  logic for allowed operator-expression and tests, Adds  paramete…

julianhille:
This pull requests fixes some issues with the not operator and is now testing the allowed operations for a $not operator.
Also i added the $eq operator and a test for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: python
-sudo: false
+sudo: true
 cache:
   directories:
     - $HOME/.cache/pip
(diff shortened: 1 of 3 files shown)


vmalloc:
@julianhille the Travis CI job is failing on this PR

mdomke:
Hmm… That is strange. The respective failing test works if I try the query manually on my machine. I suspect that the MongoDB version used in the tests on the travis server is too old (MongoDB 2.4.12).

mdomke:
I actually could confirm that. The query

```
db.collection.find({'name': {'$not': {'$eq': 'bob'}}})
```

works with the MongoDB server version on my machine (3.0.5), but fails with

```
error: { "$err" : "invalid use of $not", "code" : 13034 }
```

for a server version 2.4.x.

@julianhille: Either skip that test or require a higher MongoDB server version in the travis CI build…

vmalloc:
I think installing it as a part of the build is an acceptable practice.

mdomke:
Sure it is possible to install a more recent version of the MongoDB server, but it's still kind of hacky to fiddle around with additional apt-sources instead of just naming the dependent service and let travis deal with all the installation process. Nevertheless I'm not sure if the above mentioned PR is being closed anytime soon, which would provide the preferable clean solution. Something like in this `.travis.yml` file would probably work https://github.com/kraih/mango/blob/master/.travis.yml

mdomke:
@julianhille I didn't see that this pull request was made against the master branch so that I head to revert the PR. Could you please open a new PR against develop? Thank you!

PR #178, opened 2015-09-01
Added support for bulk operations

ddzialak:
While adding tests I've found and fixed two bugs:
- handle multi=False in db.collection.remove
- fix in update with upsert=True (extra document should not be inserted if any other modified)

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -5,6 +5,7 @@ import itertools
 import json
 import time
 import warnings
+import functools
 try:
     from collections import OrderedDict
 except ImportError:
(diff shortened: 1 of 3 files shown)


mdomke:
Thank you for contributing! I'd currently only have some comments regarding coding style and PEP-8 compliance but since the project does not have an automated style checker (yet!), I wont be that pedantic. Nevertheless would it be possible for you to write a simple back-to-back test with pymongo like it is done in the tests in `test__mongomock.py`? That would be awesome!

ddzialak:
finally... it was hard because unfortunately it occurred that {{nModified}} key sometimes is not returned and it is dependent on mongo server version. I've added one hacky solution to know if server returns {{nModified}} key or not but it's only used with {{CollectionComparisonTest} that is skipped if there is no pymongo.

mdomke:
Looks good to me now. Thank you very much!

PR #180, opened 2015-09-04
Changes setup to pbr and travis and dependency building to tox

julianhille:


mdomke:
Well I guess you were faster here than me, since I was also planing to do the packaging with pbr.

PR #187, opened 2015-09-15
backward compatibility

kfrendrich:


mdomke:
The 3.x branch of pymongo has introduced some breaking changes, including the deprecation of some parameters to functions. In my opinion it is not feasible to support both, version 2.x and 3.x of pymongo, at the same time. You couldn't rely on your mocked code to behave in the same way as the production could (e.g.: the mocked code would accept a parameter that makes your production code explode). If you actually want to use pymongo 2.x in your code, than I'd suggest to use mongomock 2.x. Just install it with 

```
pip install mongomock < 3.0
```

If you have any features you specifically want to add to the 2.x branch of mongomock make a PR against the `support/2.x` branch.

PR #188, opened 2015-09-18
Fix find's projection argument so that it works with nested fields of…

marcinbarczynski:
… any level


mdomke:
Thanks for the support! Can you please create the PR against the develop branch?

PR #199, opened 2016-01-10
Implement get_default_database (Fix #198)

### LINKED ISSUE #198: `get_default_database` is not implemented
The pymongo mongoclient method `get_default_database` is not implemented. It instead trys to return a database named "get_default_database".

The method is documented at http://api.mongodb.org/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient.get_default_database.

sm11963:
Add a simple implementation of `MongoClient.get_default_database`. It always returns the same database `__DEFAULT__` specified by the class field `MongoClient.DEFAULT_DB`. 

This would be more accurately implemented by using `pymongo.uri_parser.parse_uri` ([docs](http://api.mongodb.org/python/current/api/pymongo/uri_parser.html)) to parse the default database from the URI if the MongoClient was created with a URI and otherwise throw an error if not.

I did not use this approach because I wanted to get some feedback on the best approach for how to do that since I'm guessing we do not want to add pymongo as a dependency. 

From pymongo:
```
def get_default_database(self):
    if self.__default_database_name is None:
        raise ConfigurationError('No default database defined')

    return self[self.__default_database_name]
```

### DIFF
--- a/mongomock/__init__.py
+++ b/mongomock/__init__.py
@@ -1,5 +1,3 @@
-from .helpers import ObjectId  # noqa
-
 try:
     from pymongo.errors import PyMongoError
 except ImportError:
@@ -30,6 +28,19 @@ except ImportError:
     class InvalidOperation(PyMongoError):
         pass
 
+try:
+    from pymongo.errors import ConfigurationError
+except ImportError:
+    class ConfigurationError(PyMongoError):
+        pass
+
+try:
+    from pymongo.errors import InvalidURI
+except ImportError:
+    class InvalidURI(ConfigurationError):
+        pass
+
+from .helpers import ObjectId  # noqa
 from mongomock.__version__ import __version__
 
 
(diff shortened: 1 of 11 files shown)


mdomke:
I would actually prefer if the implementation would behave as much as possible as the pymongo code itself. That means that it should raise a ``ConfiguartionError`` in the way you described it above (if the ``MongoClient`` has not been initialised with a proper URL). I'm aware that that would imply the necessity to implement at least a simple URI parser (but maybe the ``urllib2.urlparse.urlparse`` already does most of the job).

sm11963:
@mdomke agreed. I went back through and re-implemented it. However instead of using `urllib2.urlparse.urlparse` I tried to replicated the pymongo code as much as possible without bringing over any unnecessary complexities (so all the parsing and testing only focuses on the database, the only component of the URI which is currently used). 

For the uri_spec testing I wanted to directly copy the pymongo spec tests to ensure a full spec testing (to ensure any valid URI is parsed valid) and allow easy expansion of tests in the future. Much of the test specification is unused however since only the database is parsed out.

I hope this is more appropriate. I would be happy to make any changes or additions that you see fit.

mdomke:
@sm11963 thank you very much for your effort!

PR #202, opened 2016-01-15
fix unexcepted exception when $regex meets non-string

### LINKED ISSUE #201: regex match will raise Exception when target is not a string
version:3.4

In mongodb:

```
> use test
> db.test.save({'a':1})
> db.test.find()
{ "_id" : ObjectId("569a1d46b84676ece871ad44"), "a" : 1 }
> db.test.find({'a':{'$regex':'1'}})
```

In mongomock
```
In [1]: import mongomock

In [2]: mongomock.filtering.filter_applies({"a":{"$regex":"1"}}, {"a":1})
....
/usr/local/lib/python3.5/site-packages/mongomock/filtering.py in <genexpr>(.0)
    150 
    151 def _regex(doc_val, regex):
--> 152     return any(regex.search(item) for item in _force_list(doc_val))
    153 
    154 

TypeError: expected string or bytes-like object

```

slayercat:
fix #201

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -149,6 +149,8 @@ def _elem_match_op(doc_val, query):
 
 
 def _regex(doc_val, regex):
+    if not (isinstance(doc_val, (string_types, list)) or isinstance(doc_val, RE_TYPE)):
+        return False
     return any(regex.search(item) for item in _force_list(doc_val))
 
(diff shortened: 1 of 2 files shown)


mdomke:
Thank you for the fix and sorry for the delayed response from my side!

PR #209, opened 2016-02-16
Basic support for $currentDate operation

lifus:
Timestamp is not supported so far, hence throwing `NotImplementedError` when [typeSpecification]( https://docs.mongodb.org/manual/reference/operator/update/currentDate/#definition) is exactly `{'$type': 'timestamp'}`.
Not sure if `typeSpecification` should be restricted to `True` or `{'$type': 'date'}` in mongomock. Since I've noticed some weird behavior in mongo, e.g. it allows `$currentDate: {foo: False}` and treats it exatly the same as  `$currentDate: {foo: True}` while prohibiting `$currentDate: {foo: 1}`

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -362,27 +362,14 @@ class Collection(object):
             subdocument = None
             for k, v in iteritems(document):
                 if k == '$set':
-                    positional = False
-                    for key in iterkeys(v):
-                        if '$' in key:
-                            positional = True
-                            break
-                    if positional:
-                        subdocument = self._update_document_fields_positional(
-                            existing_document, v, spec, _set_updater, subdocument)
-                        continue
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -860,6 +846,16 @@ def _update_document_fields_positional(self, doc, fields, spec, updater,
 
         return subdocument
 
+    def _update_document_fields_with_positional_awareness(self, existing_document, v, spec,
```
mdomke: Thanks for refactoring this into a method and making the code more DRY!

mdomke:
There are some merge conflicts in the `test__collection_api.py` file, after I merged your last PR. Could you please rebase this PR to the develop HEAD?

lifus:
Build failed with `gpgkeys: key 7F0CEB10 not found on keyserver`

Issue #211, opened 2016-02-22
Incorrect upsert of document with dotted attribute criteria


syev:
When updating with `upsert=True` a document using a dotted attribute as a criteria, if:
- no match is found
- the update operation is a partial update (not a replace)

Then the dotted attribute should be expanded to the corresponding subdocuments in the document to be inserted.
But currently mongomock inserts a document with the dotted attribute "as is".

**Example**
On an empty collection, the following code:
```
mongo = mongomock.MongoClient()
mongo.test.test.update_one({'a.b': 1}, {'c': 2}, upsert=True)
```
Should result in the insertion of a document like this:
```
{
  '_id': ObjectId("..."),
  'a': {
    'b': 1
  },
  'c': 2
}
```
Whereas the current mongomock implementation inserts a document like this:
```
{
  '_id': ObjectId("..."),
  'a.b': 1,
  'c': 2
}
```

PR #212, opened 2016-02-22
Issue #211: Expand dotted attributes on upsert

syev:
Fix for #211

- Expand dotted query attributes before inserting in the case of an upsert
- Raise WriteError on conflict
- Unit tests


mdomke:
Thanks!

PR #225, opened 2016-05-18
use xrange from six lib

srinivasreddy:


srinivasreddy:
@mdomke  Please take a look and merge

PR #230, opened 2016-06-11
Use dict comprehensions instead of list comprehensions

thedrow:
It's much more idiomatic and possible since we don't support Python 2.6.
It's also slightly faster.


mdomke:
Good finding!

PR #231, opened 2016-06-11
Add Python 3.5 to the build matrix

thedrow:


mdomke:
In the course of this, we should also adapt the trove classifiers in the `setup.cfg`. Could you do that?

thedrow:
Done :)

PR #232, opened 2016-06-11
Cache pip explicitly in Travis

thedrow:
It's now allowed to specify cache: pip even if you override the install section.


mdomke:
Thanks @thedrow!

PR #235, opened 2016-06-28
Add support for $min and $max operations

plocharz-9livesdata:


mdomke:
Thanks, @plocharz-9livesdata!

PR #260, opened 2016-09-20
Add aggregation project stage support

drorasaf:
Add basic projection support, add support for almost all arithmetic opeartors for expression in project, add min operator, add relevant tests

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1368,13 +1368,13 @@ class Collection(object):
 
         out_collection = [doc for doc in self.find()]
         grouped_collection = []
-        for expression in pipeline:
-            for k, v in iteritems(expression):
+        for stage in pipeline:
+            for k, v in iteritems(stage):
                 if k == '$match':
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
                 elif k == '$group':
-                    _id = expression['$group']['_id']
+                    _id = stage['$group']['_id']
                     group_func_keys = self._get_group_id_fields(_id)
 
                     out_collection, group_func_keys = self._add_group_id_fields(out_collection,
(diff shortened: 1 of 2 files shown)


drorasaf:
Maybe next step is to enable expression parsing for group stage as well

mongomock/collection.py
```diff
@@ -1363,18 +1391,102 @@ def aggregate(self, pipeline, **kwargs):
             '$hour',
             '$minute',
             '$second',
-            '$millisecond']
-        conditional_operators = ['$cond', '$ifNull']  # noqa
+            '$millisecond',
+            '$dateToString']
+
+        def _handle_arithmetic_operator(operator, values, doc_dict):
+            import math
```
srinivasreddy: pls import at the top

mongomock/collection.py
```diff
@@ -1363,18 +1391,102 @@ def aggregate(self, pipeline, **kwargs):
...
+            else:
+                raise NotImplementedError("%s is not supported operator for the aggregation "
+                                          "pipeline. See http://docs.mongodb.org/manual/meta/"
+                                          "aggregation-quick-reference/ for a complete list of "
+                                          "valid operators." % k)
+
+        def _extend_collection(out_collection, field, expression):
+            field_exists = False
+            for doc in out_collection:
+                if field in doc.keys():
```
srinivasreddy: `s/doc.keys()/doc`

mongomock/collection.py
```diff
@@ -1363,18 +1391,102 @@ def aggregate(self, pipeline, **kwargs):
...
+                    field_exists = True
```
srinivasreddy: you need to `break` after field_exists `True`

drorasaf:
@srinivasreddy  Fixed.
Added another fix: instead of getting None return NOTHING, since none might be actual value.

srinivasreddy:
Thanks

Issue #237, opened 2016-06-30
Important tests being skipped


matt-snider:
Hi! 

I'm new to the project, and was looking to make some contributions to aggregation pipeline support that I think would be quite useful. 

As I've been looking through the code, I noticed that some tests are being skipped by tox in both the pymongo and pyexecjs environments. This is due to some tests have both `skipIf()` decorators, for example:

``` python
@skipIf(not _HAVE_PYMONGO, "pymongo not installed")
@skipIf(not _HAVE_MAP_REDUCE, "execjs not installed")
class MongoClientAggregateTest(_CollectionComparisonTest):
      #....
```

It seems to me that the simplest solution would just be to modify the tox.ini deps to include pyexecjs in the pymongo environment, and add verbose logging to the nosetest call to make it easier to recognize when something is being incorrectly skipped.

I wanted to get in touch before opening a PR in case there's a reason it's done this way. :smile:

texuf:
Hey sorry nobody ever got back to you. I'm sure the tests were skipped before we even used Tox.

Were you able to get this done?

matt-snider:
Hi @texuf! No worries. 

I did some work on it but didn't get around to finishing it. I'll take a look now and possibly open a PR

matt-snider:
@texuf Finished the PR last week, but there are some issues with tests from the last few PRs failing since these tests were never actually run. They should pass if tox is run with a newer Mongo version. See #265

smalekzadeh:
Hi 
Can anyone help please? Does mongomock has support for test aggregate pipelines?
Thanks

smalekzadeh:
I get the "TypeError: 'Collection' object is not callable" error when try to use it

pcorpet:
Thanks for reporting and sorry for the long delay in the response. This was eventually fixed by ac610fc and then further down by bd99411.

PR #265, opened 2016-10-03
Fix skipping of MongoClientAggregateTest

matt-snider:
As discussed in #237, tests that require both pymongo and pyexecjs aren't being run at all since there is no tox factor which includes both. To fix this, I've just added the pyexecjs dependency to the pymongo dependencies. I'll add the commit that fixes the failing tests after the CI runs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,8 @@ env:
 script:
   - tox -e $TOX_ENV
 before_install:
-  - "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10"
-  - "echo 'deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen' | sudo tee /etc/apt/sources.list.d/mongodb.list"
+  - "sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv EA312927"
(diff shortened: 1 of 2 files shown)


matt-snider:
Hmm okay, well it looks like this change uncovered some failing tests that weren't being run. They seem to only be failing because the CI runs with an older version of mongo (2.6.12). 

All tests pass when running it locally with mongo 3.2

matt-snider:
Hi @srinivasreddy! Any comment or feedback on tis?

srinivasreddy:
@matt-snider  Thanks for the PR and apologies for the late reply.

Could you please resolve conflicts. I will take a look in a day or two.
 
Thanks for your patience.

matt-snider:
@srinivasreddy No problem. I've rebased onto develop and reverted one commit, as it was fixed in 50c694180a9d266d4a5bb3e322bb57bbf6ee8783

matt-snider:
But I guess the Travis build won't be triggered by an old commit. Can it be triggered manually?

srinivasreddy:
@matt-snider  Yes. You can do that by closing the pr and opening again. :smile:

matt-snider:
As mentioned above, the failing tests are due to the mongo version used. The $min operator for $project stages [only became available in v3.2](https://docs.mongodb.com/v3.2/reference/operator/aggregation/min/). 

I've updated the `tox.ini` file to use mongodb 3.2, but I'm not sure if this is desirable. 

Let me know what you think.

srinivasreddy:
I am not sure of this.  tagging. @mdomke  @vmalloc

pcorpet:
This was eventually fixed by #424

Issue #206, opened 2016-02-14
Using "{'$not': {'$size': 0}}" raises OperationFailure


ymost:
The following code runs fine with pymongo:

```
col.insert_many([{'a': [1,2,3]}])
cursor = col.find({'a': {'$not': {'$size': 0}}})
cursor.next()
```

But with mongomock you get:

```
OperationFailure: BadValue $not needs a regex or a document
```

dgarana:
Hi, this is related with this pull request.
https://github.com/mongomock/mongomock/pull/266

pcorpet:
This was fixed by @srinivasreddy in #266, thanks Srinivas.

PR #266, opened 2016-10-03
Add suport for $size operator

dgarana:
Add support for enabling $size operator, which counts how many
array elements are on a field:

https://docs.mongodb.com/manual/reference/operator/query/size/

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -154,6 +154,13 @@ def _regex(doc_val, regex):
     return any(regex.search(item) for item in _force_list(doc_val))
 
 
+def _size_op(doc_val, search_val):
+    if isinstance(doc_val, (list, tuple, dict)):
+        return search_val == len(doc_val)
+    else:
+        return search_val == 1 if doc_val else 0
+
+
 OPERATOR_MAP = {
     '$eq': operator.eq,
     '$ne': operator.ne,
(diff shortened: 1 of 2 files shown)


dgarana:
This PR should solve: https://github.com/mongomock/mongomock/issues/206

mongomock/filtering.py
```diff
@@ -154,6 +154,13 @@ def _regex(doc_val, regex):
     return any(regex.search(item) for item in _force_list(doc_val))
 
 
+def _size_op(doc_val, search_val):
+    if isinstance(doc_val, (list, tuple, dict)):
+        return search_val == len(doc_val)
+    else:
+        return search_val == 1 if doc_val else 0
```
dgarana: This could be changed with:
```
+def _size_op(doc_val, search_val):
+    if isinstance(doc_val, (list, tuple, dict)):
+        return search_val == len(doc_val)
+    return int(doc_val is not None)
```
If doc value is not a list, it's size is based in if it's exists or not.

srinivasreddy: i am not really comfortable in treating 1 as True, 0 as False. So   `s/return int(doc_val is not None)/return  doc_val is not None` should work.

dgarana: Somewhere on this conversation we've forget what the change is about. This change supposedly should check if an array of items is as long as we expect with $size operator.

If item isinstance of an iterable type(list, tuple, dict) it should check it's length, else, it's size is based in if it exists or not (1 or 0, that doesn't mean True or False, it is it's length).
So, I'd recommend this change:
```
+    if isinstance(doc_val, (list, tuple, dict)):
+        return search_val == len(doc_val)
+    return search_val == 1 if doc_val else 0
```

PR #270, opened 2016-10-06
Empty field value behave like null

KrissMcMacer:


srinivasreddy:
Can you give me an example for buggy behaviour?

KrissMcMacer:
You can see it in test__insert_empty_doc_uniq_idx

PR #271, opened 2016-10-06
collection: name func params same as pymongo3

jordan-heemskerk:
- Rename update_one, update_many, and replace_one function parameters to
  be consistent with pymongo3
- Update tests to use keyword args when calling these functions


srinivasreddy:
thanks

PR #272, opened 2016-10-17
fix for updating with {}

bugrimov:
When calling `update({'name': 'bob'}, {})` it must clear document like in real mongo, but now it leaves document unchanged.


srinivasreddy:
thanks

Issue #287, opened 2017-01-12
Projection does not work across list of subdocuments


Bernie:
When attempting to project a subset of fields from a list of subdocuments, MongoMock fails to properly emulate MongoDB behaviour, when occurring both in the find and aggregate use cases.

Take this test document as an example:

`{'a': 1, 'b': [{'c': 3, 'd': 4}, {'c': 5, 'd': 6}]}`

When performing a find operation in MongoDB, and projecting only one of the fields from the subdocument, for instance, field 'b.c', MongoDB returns field 'c' across all subdocuments in the list:

`find_one({'a': 1}, {'b.c': 1})`:

`{'_id': ObjectId('587795f7b0c3993ebd9219fb'), 'b': [{'c': 3}, {'c': 5}]}`

The same behaviour applies when projecting in an aggregation:

`aggregate([{'$match': {'a': 1}}, {'$project': {'b.c': 1}}])`:

`{'_id': ObjectId('587795f7b0c3993ebd9219fb'), 'b': [{'c': 3}, {'c': 5}]}`

However, this breaks in MongoMock.

When performed during the find operation, an empty dict is returned instead of a list of projected subdocuments:

`{'b': {}, '_id': ObjectId('5877922eb0c3993dcebdf17d')}`

When the same projection is applied in the aggregation, the aggregation operation fails with an exception:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mongomock/collection.py", line 1586, in aggregate
    out_collection = _extend_collection(out_collection, field, value)
  File "mongomock/collection.py", line 1464, in _extend_collection
    doc[field] = _parse_expression(expression.copy(), doc)
AttributeError: 'int' object has no attribute 'copy'
```

Bernie:
I have a fix for this, however it does not match MongoDB behaviour exactly when projections are requests that do not match the documents structure.

However, it appears to behave more closely to MongoDB than before, as it now supports list of subdocuments properly.

PR #288, opened 2017-01-12
Fix projection

Bernie:
Fixes projection of subdocument lists.

There is still some difference with MongoDB when specifying projections that do not match the document's structure.

Fix for issue #287

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -798,7 +798,9 @@ class Collection(object):
     def _subdocitem_merge(self, doc_copy, field, sdocitems, subtract=False):
         sdoc_copy = doc_copy.setdefault(field, [])
         if len(sdoc_copy) < len(sdocitems):
-          sdoc_copy.extend([{} if not subtract else {'$EMPTY$': True} for _ in xrange(len(sdocitems)-len(sdoc_copy))])
+            sdoc_copy.extend([{} if not subtract else {'$EMPTY$': True}
+                for _ in xrange(len(sdocitems)-len(sdoc_copy))
+            ])
         for (sdoc, sdocitem) in zip(sdoc_copy, sdocitems):
             if not subtract:
                 sdoc.update(sdocitem)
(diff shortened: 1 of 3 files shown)


mongomock/collection.py
```diff
@@ -826,6 +795,77 @@ def _copy_only_fields(self, doc, fields, container):
...
+                if sdoc == {'$EMPTY$': True}:
+                    sdoc.clear()
+                    sdoc.update(sdocitem)
+                else:
+                    exc_keys = set(sdoc.keys()).difference(sdocitem.keys())
+                    for key in exc_keys:
+                        del sdoc[key]
+
+    def _filter_subdocuments(self, doc, fields, container):
+        # if 1 was passed in as the field values, include those fields
```
mdomke: Comments above an `if` statement, where each suite holds a big chunk of code usually indicates a code-smell, where each suite should be a separate function or method. In this case additionally both suites look very similar (a.k.a copy-pasted), so that I think they should be refactored to a function. I know that you just adapted the code that was there before, but it would be nice if you could remove the obvious flaws while adding functionality.

mongomock/collection.py
```diff
@@ -826,6 +795,77 @@ def _copy_only_fields(self, doc, fields, container):
...
+        if list(fields.values())[0] == 1:
+            doc_copy = container()
+            for key in fields:
+                key_parts = key.split('.')
+                subdocument = doc
+                subdocument_copy = doc_copy
+                full_key_path_found = True
+                for (i, key_part) in enumerate(key_parts[:-1]):
```
mdomke: I find it very uncommon to put parentheses around the target-list of the for-loop-statement. Even though that it is syntactically and semantically equivalent most people will write
```python
for i, item in enumerate(iterable):
    ...
```

mongomock/collection.py
Bernie: I think you're expecting a bit much from a volunteer.

I'm more than happy to fix bugs or extend functionality that I've found useful for my own purposes, but to expect me to rewrite a portion of the existing code because the original code was poorly written seems a bit unfair.

pcorpet:
#331 has been submitted and cover that code already, so I'm closing this. Please reopen if I'm missing something.

Issue #290, opened 2017-01-20
Document fetched from collection is actual reference and not a copy of the document


Bernie:
Documents returned from MongoMock should be copies of the documents stored in the collection in order to properly emulate a MongoDB.

When a document fetched from a MongoDB is modified, no changes are made to the DB unless an explicit call to save or update is made.

In MongoMock, because the document returned from the fetch operation is the actual document and not a copy, a modification of the document results in the document in the collection being immediately modified.  This is not desirable when emulating MongoDB.

PR #291, opened 2017-01-20
Return the copy of the document, not the actual document.

Bernie:
This only addresses the problem when fetching a document from the collection or through a find operation.

I have not checked when fetching a document through the aggregation pipeline.

This addresses issue #290.


srinivasreddy:
Thanks!!

PR #292, opened 2017-01-25
store index information and return index name from create_index

BeardedSteve:
This allows you to test if index creation was performed as expected on collections.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -385,7 +385,8 @@ class Collection(object):
         self._documents = OrderedDict()
         self._force_created = create
         self._write_concern = write_concern or WriteConcern()
-        self._uniques = []
+        self._uniques = {}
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -1062,14 +1064,24 @@ def ensure_index(self, key_or_list, cache_for=300, **kwargs):
...
+
+
+        return index_string
 
     def drop_index(self, index_or_name):
         pass
 
     def index_information(self):
-        return {}
+        return self._index_information
```
pcorpet: Please return a copy of the dict, to make sure it's immutable (i.e. if the caller modifies the return parameter it does not change the value of the next call).

pcorpet:
So I went ahead and did it myself. I'm going to merge your PR. Thanks for the contribution.

BeardedSteve:
thank you @pcorpet - sorry I couldn't help out this time I'm travelling at the moment.

PR #293, opened 2017-01-27
Add a data persistence layer to mongomock

MacHu-GWU:
Add a feature that can dump and load database. So user can use mongodb
as a pure python, local database. And user can easily switch to real
mongodb by changing the MongoClient setting.


MacHu-GWU:
Hi, since mongomock can act like pymongo, why not adding a data persistence layer to it? Even the performance is not really good. But for prototype dev, it's usefule. Thank you.

PR #297, opened 2017-02-27
fix operator filtering inside list fields

OriHoch:
#### reproduction steps
* insert mongo documents containing a list field, e.g.:
```
{
    "years": [1956, 1978, 1987]
}
```
* try to find this document using $gt / $lt (see https://docs.mongodb.com/manual/tutorial/query-arrays/#query-an-array)
```
db.collection.find({
    "years": {
        "$gt": 1950,
        "$lt": 1960
    }
})
```

##### expected
* mongomock should return all documents that match the operators inside the list fields

##### actual
* mongomock doesn't search inside the list - but compares the list itself

#### notes
* I assume this logic should be added to other operators (e.g. $eq / $ne) as well, but seems like it requires more complicated logic and I also don't need that functionality so I won't implement it :)


pcorpet:
Thanks Ori!

PR #309, opened 2017-03-14
Fix cursor.__getitem__ behavior

touilleMan:
According to pymongo documentation (http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.__getitem__) `Cursor.__getitem__` should update itself when passing a slice instead of returning a list of results (current behavior).

This PR fix this.


srinivasreddy:
Thanks

PR #320, opened 2017-04-04
enable tox test using both pyexecjs & pymongo

DeoLeung:


mongomock/collection.py
```diff
@@ -1534,7 +1534,7 @@ def _extend_collection(out_collection, field, expression):
                     _id = stage['$group']['_id']
                     if _id:
                         key_getter = functools.partial(_parse_expression, _id)
-                        out_collection = sorted(out_collection, key=key_getter)
+                        out_collection = sorted(out_collection, key=lambda x: str(key_getter(x)))
```
vmalloc: This change doesn't seem related to the PR's title. Why was this introduced?

tests/test__mongomock.py
```diff
@@ -1680,7 +1681,11 @@ def test__aggregate17(self):
         pipeline = [
             {'$project': {'_id': 0, 'created': {'$subtract': [{'$min': ['$a', '$b']}, '$count']}}}
         ]
-        self.cmp.compare.aggregate(pipeline)
+        # TODO(Metrodatateam): use skipif if version is accessible from decorator
```
vmalloc: ?

mongomock/collection.py
DeoLeung: hmmm, that was quite an old pr I can't fully recall why i did this...looks like I was adding this to make the test pass(it was broken)

tests/test__mongomock.py
DeoLeung: some aggregate was not supported until certain version, so I think we may better just skip some tests if it's not supported

vmalloc: Ok, so can we please replace this TODO with an open issue?

pcorpet:
@DeoLeung thanks for your help. I was able to commit a similar commit in #424.

PR #322, opened 2017-04-04
Fix issue with skipping twice in find

### LINKED ISSUE #321: Bug when using skip in find
2636fbb introduces a double 'skip', once in the dataset creation and once in the iteration.
We only need one, and for count to work (which ignores these params) we need it to be outside the dataset creation.

adamhicks:
41235a1 introduces a double 'skip', once in the dataset creation and once in the iteration.
We only need one, and for count to work (which ignores these params) we need it to be outside the dataset creation.


srinivasreddy:
Thanks

PR #331, opened 2017-05-03
Fix projection with subdocument arrays

edengreen:
Make sure that when multiple fields of a subdocument are properly projected from an array as well.

Note:
Only after writing most of the code, I came across this pull-request: #288, which is supposed to fix some, but not all of the issues I addressed.
I adapted the unit-tests that were added in that pull-request, as they were more comprehensive than mine. Hope that's ok.

This is another proposed fix for #287.


mongomock/collection.py
```diff
@@ -963,6 +926,59 @@ def _copy_only_fields(self, doc, fields, container):
                 fields[field] = op
             return doc_copy
 
+    def _combine_projection_spec(self, projection_fields_spec):
```
pcorpet: You don't use `self` here so it's better to make it a function rather than a method.

mongomock/collection.py
```diff
@@ -963,6 +926,59 @@ def _copy_only_fields(self, doc, fields, container):
...
+        """
+
+        tmp_spec = defaultdict(dict)
+        for f, v in iteritems(projection_fields_spec):
+            if '.' not in f:
+                tmp_spec.setdefault(f, v)
+            else:
+                split_field = f.split('.', 1)
+                base_field, new_field = tuple(split_field)
+                if base_field in tmp_spec and not isinstance(tmp_spec[base_field], dict):
```
pcorpet: A bit complex, `not isinstance(tmp_spec.get(base_field), dict)` is shorter.

mongomock/collection.py
```diff
@@ -963,6 +926,59 @@ def _copy_only_fields(self, doc, fields, container):
...
+                    tmp_spec[base_field] = {new_field: v}
+                else:
+                    tmp_spec[base_field][new_field] = v
+
+        combined_spec = {}
+        for f, v in iteritems(tmp_spec):
+            if not isinstance(v, dict):
```
pcorpet: In general when you do:
```py
if not a:
  statement_1
else:
  statement_2
```
I think it's easier to write:
```py
if a:
  statement_2
else:
  statement_1
```

mongomock/collection.py
```diff
@@ -963,6 +926,59 @@ def _copy_only_fields(self, doc, fields, container):
...
+        doc_copy = container()
+
+        if not is_include:
+            # copy only scalar values
+            for key, val in iteritems(doc):
+                if not isinstance(val, (list, tuple, dict)):
+                    doc_copy[key] = val
+
+        for key, spec in iteritems(combined_projection_spec):
+            if key in doc:
```
pcorpet: I prefer to `continue` early rather than wrapping the whole loop in a branch

tests/test__mongomock.py
```diff
@@ -735,6 +735,18 @@ def test__find_limit(self):
         # pymongo limit defaults to 0, returning everything
         self.cmp.compare.find(limit=0, sort=[("a", 1), ("b", -1)])
 
+    def test__find_projection_subdocument_lists(self):
+        self.cmp.do.remove()
+        self.cmp.do.insert({'a': 1, 'b': [{'c': 3, 'd': 4}, {'c': 5, 'd': 6}]})
+        for project in ({'_id': 0, 'a': 1, 'b': 1},
```
pcorpet: It's not much more work to write `self.cmp.compare.find_one({'a': 1}, ` before each projection case, and it makes potential error easier to read (otherwise, one does not know which projection broke the test).

PR #340, opened 2017-06-18
Add missing feature in map_reduce()

pcart-grandjean:
The argument _scope_, for example, is very useful to add extra elements in the scope of the map and reduce functions.


pcart-grandjean:
I only updated a rest file. How could it fail a check?

pcorpet:
My bad your commit was pretty clear. I think I got confused between "finalize the mapreduce implementation" and "the `finalize` option of mapreduce is missing" :-)

I've rebased your branch and clarified the point above.

PR #343, opened 2017-07-04
Manage dictionaries in aggregate addToSet

JulieRossi:
**What this PR does**  
This PR allows to use use $addToSet operator in aggregate on dictionaries.

**Example**

```
db.coll.aggregate([{"$group": {"_id": "$a", "set": {"$addToSet": "$dictionary"}}}])
```

**Notes**  
The corresponding tests have been added.  
I needed this functionality for my project, I hope you will find it usefull.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1766,7 +1766,7 @@ class Collection(object):
                                     elif operator == "$addToSet":
                                         doc_dict[field] = []
                                         val_it = (val or None for val in values)
-                                        # don't use set in case elt in not hashable (like dicts)
+                                        # Don't use set in case elt in not hashable (like dicts).
                                         for elt in val_it:
                                             if elt not in doc_dict[field]:
                                                 doc_dict[field].append(elt)
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1771,7 +1771,6 @@ class Collection(object):
                                             if elt not in value:
                                                 value.append(elt)
                                         doc_dict[field] = value
-                                        doc_dict[field].reverse()
                                     elif operator == '$push':
                                         if field not in doc_dict:
                                             doc_dict[field] = []
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1764,8 +1764,13 @@ class Collection(object):
                                     elif operator == "$last":
                                         doc_dict[field] = values[-1]
                                     elif operator == "$addToSet":
+                                        doc_dict[field] = []
                                         val_it = (val or None for val in values)
-                                        doc_dict[field] = set(val_it)
+                                        # don't use set in case elt in not hashable (like dicts)
+                                        for elt in val_it:
+                                            if elt not in doc_dict[field]:
+                                                doc_dict[field].append(elt)
+                                        doc_dict[field].reverse()
                                     elif operator == '$push':
                                         if field not in doc_dict:
                                             doc_dict[field] = []
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1764,12 +1764,13 @@ class Collection(object):
                                     elif operator == "$last":
                                         doc_dict[field] = values[-1]
                                     elif operator == "$addToSet":
-                                        doc_dict[field] = []
+                                        value = []
                                         val_it = (val or None for val in values)
                                         # Don't use set in case elt in not hashable (like dicts).
                                         for elt in val_it:
-                                            if elt not in doc_dict[field]:
-                                                doc_dict[field].append(elt)
+                                            if elt not in value:
+                                                value.append(elt)
+                                        doc_dict[field] = value
                                         doc_dict[field].reverse()
                                     elif operator == '$push':
                                         if field not in doc_dict:
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -1907,7 +1907,18 @@ class MongoClientAggregateTest(_CollectionComparisonTest):
         pipeline = [
             {'$group': {"_id": "$count", "set": {"$addToSet": {"a": "$a", "b": "$b"}}}},
         ]
-        self.cmp.compare_ignore_order.aggregate(pipeline)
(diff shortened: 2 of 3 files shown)


mongomock/collection.py
```diff
@@ -1570,8 +1570,13 @@ def _extend_collection(out_collection, field, expression):
                                     elif operator == "$last":
                                         doc_dict[field] = values[-1]
                                     elif operator == "$addToSet":
+                                        doc_dict[field] = []
```
pcorpet: I suggest you use a local var to build the list and only assign it to `doc_dict[field]`. This will avoid always getting the field.

mongomock/collection.py
```diff
@@ -1570,8 +1570,13 @@ def _extend_collection(out_collection, field, expression):
...
                                         val_it = (val or None for val in values)
-                                        doc_dict[field] = set(val_it)
+                                        # don't use set in case elt in not hashable (like dicts)
```
pcorpet: Please start your comments with uppercase and end them with a dot.

mongomock/collection.py
```diff
@@ -1570,8 +1570,13 @@ def _extend_collection(out_collection, field, expression):
...
+                                        for elt in val_it:
+                                            if elt not in doc_dict[field]:
+                                                doc_dict[field].append(elt)
+                                        doc_dict[field].reverse()
```
pcorpet: I'm not sure why you are reversing it: according to the [doc](https://docs.mongodb.com/manual/reference/operator/aggregation/addToSet/), the order of the elements in the output array is unspecified.

tests/test__mongomock.py
```diff
@@ -1743,6 +1743,46 @@ def test__aggregate26(self):
...
+        # group addToSet
+        pipeline = [
+            {"$group": {"_id": "$a", "nb": {"$addToSet": "$count"}}},
+            {"$sort": {"_id": 1}}
+        ]
+        # self.cmp.compare cannot be used as addToSet returns elements in an unpredictable order
+        aggregations = self.cmp.do.aggregate(pipeline)
+        expected = list(aggregations["real"])
+        result = list(aggregations["fake"])
+        assert len(result) == len(expected)
```
pcorpet: please use self.assertEqual instead of assert. Same below.

tests/test__mongomock.py
```diff
@@ -1743,6 +1743,46 @@ def test__aggregate26(self):
...
+        for i, elt in enumerate(result):
+            for k, v in elt.iteritems():
+                if type(v) is list:
+                    assert set(v) == set(result[i][k])
```
pcorpet: It seems that you're comparing result ... and result! I think you meant to check expected. I suggest that you use `zip` instead of accessing by element:
```py
for expected_elt, result_elt in zip(expected, result):
  self.assertEqual(expected_elt.keys(), result_elt.keys())
  for key in result_elt:
    if isinstance(result_elt[key], list):
      self.assertCountEqual(result_elt[key], expected_elt[key], msg=key)
    else:
      self.assertEqual(result_elt[key], expected_elt[key], msg=key)
```

tests/test__mongomock.py
```diff
@@ -1743,6 +1743,46 @@ def test__aggregate26(self):
...
+    def test__aggregate28(self):
+        # group addToSet dict element
+        self.cmp.do.remove()
+        data = [
+            {"a": {"c": "1", "d": 1}, "b": {"c": "2", "d": 2}, "nb": 1},
+            {"a": {"c": "3", "d": 3}, "b": {"c": "4", "d": 4}, "nb": 1},
+            {"a": {"c": "5", "d": 1}, "b": {"c": "6", "d": 6}, "nb": 2}
+        ]
+        for item in data:
+            self.cmp.do.insert(item)
```
pcorpet: Would it be possible to use `self.cmp.do.insert_many(data)` instead please?

tests/test__mongomock.py
```diff
@@ -1743,6 +1743,46 @@ def test__aggregate26(self):
...
+                else:
+                    assert v == result[i][k]
+
+    def test__aggregate28(self):
+        # group addToSet dict element
+        self.cmp.do.remove()
+        data = [
+            {"a": {"c": "1", "d": 1}, "b": {"c": "2", "d": 2}, "nb": 1},
+            {"a": {"c": "3", "d": 3}, "b": {"c": "4", "d": 4}, "nb": 1},
+            {"a": {"c": "5", "d": 1}, "b": {"c": "6", "d": 6}, "nb": 2}
```
pcorpet: There's not much of a case here: make sure to add at least two items that have the same "a.c", a different "a.d" but the same "b" entirely.

JulieRossi:
Hello @pcorpet 
I submitted this PR a year ago, I probably should have closed it, sorry. 
I still can update it if you are interested in these changes but I don't really need them anymore.
Let me know if you prefer me to update or close this PR

pcorpet:
Sorry it took us so long to review your code. I would be grateful if you could update it so that we could add the feature to mongomock

mongomock/collection.py
JulieRossi: I am not sure I understand exactly what you mean. Do you want me to update the rest of the code too (all previous elifs) ?

mongomock/collection.py
JulieRossi: Actually, if you remove this line, the test fails. It seems the order may not be guarantied but this must be how it is implemented. I can remove this line and update the test (check that ensembles are equal and so don't take the order into account). WDYT ?

mongomock/collection.py
pcorpet: No, just for yours:

```py
value = []
...
value.append(elt)
...
value.reverse()
...
doc_dict[field] = value
```

mongomock/collection.py
pcorpet: Exactly, I like your second solution: update the test to match the expectation of MongoDB instead of having extra expectations for Mongomock.

PR #345, opened 2017-07-27
Add few missing options/functions needed by pymodm 0.4.0

agrenott:
collation and read_concern options are just added into the interfaces, without any implementation.
create_indexes is a simple loop wrapper over create_index.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -24,13 +24,18 @@ except ImportError:
     execjs = None
 
 try:
+    from pymongo.operations import IndexModel
     from pymongo import ReadPreference
     from pymongo import ReturnDocument
     _READ_PREFERENCE_PRIMARY = ReadPreference.PRIMARY
 except ImportError:
+    class IndexModel(object):
+        pass
+
     class ReturnDocument(object):
         BEFORE = False
         AFTER = True
+
     _READ_PREFERENCE_PRIMARY = None
 
 from sentinels import NOTHING
@@ -1215,8 +1220,12 @@ class Collection(object):
         validate_is_mapping('filter', filter)
         return DeleteResult(self._delete(filter, session=session), True)
 
-    def delete_many(self, filter, session=None):
+    def delete_many(self, filter, collation=None, session=None):
         validate_is_mapping('filter', filter)
+        if collation:
+            raise NotImplementedError(
+                'The collation argument of delete_many is valid but has not been '
+                'implemented in mongomock yet')
         return DeleteResult(self._delete(filter, multi=True, session=session), True)
 
     def _delete(self, filter, multi=False, session=None):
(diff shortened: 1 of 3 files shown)


mongomock/collection.py
```diff
@@ -1146,7 +1146,7 @@ def delete_one(self, filter, session=None):
         validate_is_mapping('filter', filter)
         return DeleteResult(self._delete(filter), True)
 
-    def delete_many(self, filter, session=None):
+    def delete_many(self, filter, collation=None, session=None):
         validate_is_mapping('filter', filter)
         return DeleteResult(self._delete(filter, multi=True), True)
```
pcorpet: According to our guidelines, we don't want to have features silently ignored: if `mongomock` does not handle a case it should says so explicitly. Here that would mean to raise a `NotImplementedError` indicating that collation is a valid keyword but that Mongomock does not handle it yet. Of course you should raise that exception only if the collation parameter would imply a different behaviour (i.e. it's not None).

Note that this was not done for the `session` parameter which is a bug.

mongomock/database.py
```diff
@@ -43,7 +43,7 @@ def collection_names(self, include_system_collections=True):
         return result
 
     def get_collection(self, name, codec_options=None, read_preference=None,
-                       write_concern=None):
```
pcorpet, line 52: Same here.

tests/test__collection_api.py
```diff
@@ -377,6 +381,10 @@ def test__find_removed_and_changed_options(self):
             with self.assertRaises(TypeError):
                 self.db.collection.find({}, **option)
 
+    def test__find_collation_option(self):
```
pcorpet: I think you can drop this test, as the collation param was added to find while your PR was waiting for a review.

tests/test__collection_api.py
```diff
@@ -863,6 +875,23 @@ def test__create_uniq_idxs_without_ordering(self):
...
+    @skipIf(not _HAVE_PYMONGO, "pymongo not installed")
+    def test__create_indexes_uniq_idxs(self):
+        indexes = [
+            pymongo.operations.IndexModel([('value', pymongo.ASCENDING)], unique=True),
+            pymongo.operations.IndexModel([('name', pymongo.ASCENDING)], unique=True)
+        ]
+        self.db.collection.create_indexes(indexes)
+
+        self.db.collection.insert({"value": 1, "name": "bob"})
+        # Ensure both uniq indexes have been created
```
pcorpet: typo: uniq -> unique. Also please use the opportunity to end your comment with a dot.

tests/test__collection_api.py
```diff
@@ -863,6 +875,23 @@ def test__create_uniq_idxs_without_ordering(self):
...
 
+    @skipIf(not _HAVE_PYMONGO, "pymongo not installed")
+    def test__create_indexes_uniq_idxs(self):
+        indexes = [
+            pymongo.operations.IndexModel([('value', pymongo.ASCENDING)], unique=True),
+            pymongo.operations.IndexModel([('name', pymongo.ASCENDING)], unique=True)
+        ]
+        self.db.collection.create_indexes(indexes)
+
+        self.db.collection.insert({"value": 1, "name": "bob"})
```
pcorpet: Please use `insert_one` here and below.

pcorpet:
@agrenott Thanks for your proposal, I've applied some changes to cover my remarks. Please open another bug if it's still not compatible with pymodm.

Issue #95, opened 2014-05-11
$rename operator is not supported


baruchoxman:
$rename operator support should be implemented, as described in:
http://docs.mongodb.org/manual/reference/operator/update/rename/

mpenkov:
https://github.com/mongomock/mongomock/pull/347

PR #347, opened 2017-08-04
implement support for the $rename operator

mpenkov:
We need to support [this operation](https://docs.mongodb.com/manual/reference/operator/update/rename/).

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -416,6 +416,11 @@ class Collection(object):
                     subdocument = self._update_document_fields_with_positional_awareness(
                         existing_document, v, spec, updater, subdocument)
 
+                elif k == '$rename':
+                    for src, dst in iteritems(v):
+                        if self._has_key(existing_document, src):
+                            existing_document[dst] = existing_document.pop(src)
+
                 elif k == '$setOnInsert':
                     if not was_insert:
                         continue
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -419,7 +419,10 @@ class Collection(object):
                 elif k == '$rename':
                     for src, dst in iteritems(v):
                         if '.' in src or '.' in dst:
-                            raise NotImplementedError('renaming with dots not supported yet')
+                            raise NotImplementedError(
+                                'Using the $rename operator with dots is a valid MongoDB '
+                                'operation, but it is not yet supported by mongomock'
+                            )
                         if self._has_key(existing_document, src):
                             existing_document[dst] = existing_document.pop(src)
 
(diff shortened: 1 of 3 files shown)


mpenkov:
This implements Issue #95

mpenkov:
The Travis checks failed trying to connect to MongoDB at the very start of the build:

```$ until nc -z localhost 27017; do echo Waiting for MongoDB; sleep 1; done
Waiting for MongoDB```

I'm not sure, but I don't think my changes caused this.  Could you please double-check?

Also @vmalloc if you're still looking for maintainers, I'd like to help out.

**Comment by ns-mnamburi** [comment]
@mpenkov can you pull the new sources and run tests. It should be passing now.

**Comment by mpenkov** [comment]
@ns-mnamburi Thanks, everything builds correctly now!

**Comment by pcorpet** [comment]
Thanks Michael. For new features we prefer to also add a test in `test_mongomock.py` to compare the behavior against `pymongo` + a MongoDB server. Can you add one please?

**Comment by mpenkov** [comment]
@pcorpet I added a test. I'm not very familiar with the layout of your tests, so I'm not sure if I did it right. Could you please have a look?

File: mongomock/collection.py [hunk 1]
```diff
@@ -416,6 +416,11 @@ def _update(self, spec, document, upsert=False, manipulate=False,
                     subdocument = self._update_document_fields_with_positional_awareness(
                         existing_document, v, spec, updater, subdocument)
 
+                elif k == '$rename':
+                    for src, dst in iteritems(v):
+                        if self._has_key(existing_document, src):
```
pcorpet: I think you don't handle the case where `src` or `dst` contains a dot. You can either try to implement it or check if `src` or `dst` contains a dot and raise a `NotImplementedError` if it's the case, explaining that this is a valid operation but that `mongomock` does not handle it yet.

pcorpet:
The test is perfect. Please handle the corner case before I can merge your PR.

mongomock/collection.py
```diff
@@ -416,6 +416,13 @@ def _update(self, spec, document, upsert=False, manipulate=False,
                     subdocument = self._update_document_fields_with_positional_awareness(
                         existing_document, v, spec, updater, subdocument)
 
+                elif k == '$rename':
+                    for src, dst in iteritems(v):
+                        if '.' in src or '.' in dst:
+                            raise NotImplementedError('renaming with dots not supported yet')
```
pcorpet: Make sure to identify `mongomock` explicitly as the culprit. e.g. "Using the $rename operator with dots is a valid MongoDB operation, however it is not supported by mongomock yet".

PR #353, opened 2017-10-06
Add a missing comma.

Denaun:
Was causing both $lookup and $out to appear as invalid MongoDB commands,
instead of not implemented.


vmalloc:
Thanks!

Issue #356, opened 2017-10-20
find_and_modify accepts projection kwarg instead of fields


petee-d:
`find_and_modify` method should be accepting kwarg `fields` to specify the projection (https://docs.mongodb.com/manual/reference/method/db.collection.findAndModify/), instead in mongomock it accepts kwarg `projection` for that purpose, similar to the new `find_one_and_*` methods.

Since this lead to bugs in my project, in my fix PR I'll include a commit that will raise TypeError if projection kwarg is supplied, to help discover the true purpose of the bugs.

PR #357, opened 2017-10-20
Fix find and modify projection kwarg name

petee-d:
Fixes https://github.com/mongomock/mongomock/issues/356

If the TypeError is overkill, it can be removed from the PR, but I think it's better to let people know if they use find_and_modify in a wrong way explictly, instead of strange bugs in code using mongomock the wrong way no longer respecting projection.


vmalloc:
Thanks!

PR #359, opened 2017-10-25
Add 'With ... as' statement support

Celend:
Add `__enter__` and `__exit__` methods to support 'with ... as' statement for result cursor.


vmalloc:
@Celend Where is this behavior documented in pymongo?

vmalloc:
@Celend cool! Thanks!

Issue #348, opened 2017-08-28
push_all raises ValueError


nadavperetz:
When performing:

```
from mongoengine import Document, fields, register_connection, StringField

register_connection(alias='front_end', name='front_end', is_mock=True,
                    host=getattr(settings, 'MONGO_URI'), authentication_source='admin')


class PartnerTags(Document):
    partner_uuid = fields.StringField()
    key_values = fields.ListField(StringField())

    meta = {'db_alias': 'front_end', 'collection': 'partner_tags', 'strict': False}


PartnerTags.objects(partner_uuid=partner_uuid).update_one(upsert=True, **{
            'push_all__key_values': ["1", "2", "3"]
        })
```

Raises:

```
File"..python3.5/site-packages/mongomock/collection.py", line 600, in _update
    raise ValueError('field names cannot start with $ [{}]'.format(k))
ValueError: field names cannot start with $ [$pushAll]
```

Python 3.5
mongoengine 0.11/0.13/0.14 

Thanks

uelei:
I'm doing a PR for that ! 
can you try install from my repo and run your tests ! 

git+git://github.com/uelei/mongomock@pushall_agg

nadavperetz:
It worked!

pcorpet:
A loooooong time later: pushAll is actually deprecated in MongoDB [since v2.4](https://docs.mongodb.com/v3.2/reference/operator/update/pushAll/) and completely gone in MongoDB after v3.5. Mongoengine fixed that bug here: https://github.com/MongoEngine/mongoengine/pull/1737. So I think we can close this issue.

PR #361, opened 2017-11-15
add pushall to aggregation

uelei:
related to 
https://github.com/mongomock/mongomock/issues/348

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -596,6 +596,59 @@ class Collection(object):
                                 push_results.append(value)
 
                             subdocument[nested_field_list[-1]] = push_results
+                elif k == '$pushAll':
+                    for field, value in iteritems(v):
+                        nested_field_list = field.rsplit('.')
+                        if len(nested_field_list) == 1:
+                            if field not in existing_document:
+                                existing_document[field] = []
+                            if isinstance(value, list):
+                                existing_document[field].extend(value)
+                            else:
+                                existing_document[field].append(value)
+                            continue
+                        # nested fields includes a positional element
+                        # need to find that element
+                        elif '$' in nested_field_list:
+                            if not subdocument:
+                                subdocument = self._get_subdocument(
+                                    existing_document, spec, nested_field_list)
+
+                            # we're pushing a list
+                            push_results = []
+                            if nested_field_list[-1] in subdocument:
+                                # if the list exists, then use that list
+                                push_results = subdocument[nested_field_list[-1]]
+
+                            if isinstance(value, list):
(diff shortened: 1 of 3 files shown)


tests/test__collection_api.py
```diff
@@ -1406,3 +1406,87 @@ def test_aggregate_unwind_push_first(self):
                 '_id': 'ooo'
             }]
         self.assertEqual(expect, list(actual))
+
+    def test_aggregate_unwind_pushall(self):
```
vmalloc: Please try to compact this code, it unnecessarily spans across a lot of lines...

tests/test__collection_api.py
```diff
@@ -1406,3 +1406,87 @@ def test_aggregate_unwind_push_first(self):
...
+                        }
+                    ],
+                    'id': 'ooo',
+                    'update_time': 1222
+                }
+            ]
+        )
+        actual = collection.aggregate(
+            [
+                {'$sort': {'update_time': -1}},
```
pcorpet, line 1458: This is unrelated to the feature you're trying to test. I would prefer to have unit tests specific to the feature that they are testing: if the test fails it's easier to understand what feature is causing it.

If you really want to test this whole features together, please create a separate test for it that you name accordingly.

mongomock/collection.py
```diff
@@ -1576,6 +1631,10 @@ def _extend_collection(out_collection, field, expression):
                                         if field not in doc_dict:
                                             doc_dict[field] = []
                                         doc_dict[field].extend(values)
+                                    elif operator == '$pushAll':
```
pcorpet, line 1634: It looks like the same implementation than the $push operator above: I think one of them is wrong. Let's fix this.

mongomock/collection.py
```diff
@@ -595,6 +595,59 @@ def _update(self, spec, document, upsert=False, manipulate=False,
                             else:
                                 push_results.append(value)
 
+                            subdocument[nested_field_list[-1]] = push_results
+                elif k == '$pushAll':
+                    for field, value in iteritems(v):
```
pcorpet, line 600: I have the feeling that most of this could be factorized with the $push implementation.

mongomock/collection.py
pcorpet: After checking, pushAll is not a valid group operator.
https://docs.mongodb.com/manual/reference/operator/aggregation/group/#pipe._S_group

pcorpet:
After digging some more in #348, I've decided we won't implement pushAll in mongomock. Sorry @uelei for letting your work wait for so long and reach that end.

PR #362, opened 2017-11-21
Fix behaviour of projection values in collection's find method.

pcorpet:
This avoids confusion for users who might use
```py
db.collection.find({}, {
  'myField': '1',
})
```

See the use of `'1'` instead of `1`: the mongomock library would have handled `'1'` as `0` in that case!

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -808,8 +808,9 @@ class Collection(object):
                     doc_copy = container()
                 else:
                     doc_copy = self._copy_field(doc, container)
-            # if 1 was passed in as the field values, include those fields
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -809,7 +809,7 @@ def _copy_only_fields(self, doc, fields, container):
                 else:
                     doc_copy = self._copy_field(doc, container)
             # if 1 was passed in as the field values, include those fields
```
jheden: Might want to change this to "if a truthy value", to reflect the change.

pcorpet: Done. Thanks.

PR #367, opened 2017-12-22
Make distinct() works with list of nested fields

audricschiltknecht:
Currently, calling distinct on a field that is a list of subdocuments only resolve distinct items of the first document.
Try to iterate through all of the values and correctly return the proper expected result.

This commit passes the tests (from what I can tell as some are failing on master branch) and works for my use case, but I would be interested in getting comments/critics from other people.


pcorpet:
@audricschiltknecht thanks for your proposal and sorry for not reviewing it earlier. I've rebased your code and added some tests. I would really like to add some tests in test__monogmock that cheks the differences between pymongo + MongoDb and mongomock, however for now it's not trivial as the result of distinct can have any order. I'll try to find a fix soonish.

PR #370, opened 2018-01-05
ADD : Simple aggregate operations for $project

fpozzobon:
Hi all,

Please correct me if I'm wrong but I think $min / $max / $avg / $sum operations behave the same for $group and $project.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -42,7 +42,6 @@ from sentinels import NOTHING
 from six import iteritems
 from six import iterkeys
 from six import itervalues
-from six import MAXSIZE
 from six import string_types
 from six import text_type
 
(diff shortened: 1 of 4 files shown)


mongomock/collection.py
```diff
@@ -255,15 +255,15 @@ def execute(self, write_concern=None):
     def add_insert(self, doc):
         self.insert(doc)
 
-    def add_update(self, selector, doc, multi, upsert, collation=None):
+    def add_update(self, selector, doc, multi, upsert, collation=None, array_filters=None):
```
pcorpet: As explained in the mongomock doc:
>  In such cases it is better to throw NotImplementedError than implement a modified version of the original behavior.

Please raise a `NotImplementedError` if array_filters is not `None`.

tests/test__mongomock.py
```diff
@@ -1704,6 +1704,14 @@ def test__aggregate20(self):
         ]
         self.cmp.compare.aggregate(pipeline)
 
+    def test__aggregate20b(self):
```
pcorpet: Please continue counting and move it at the bottom of the tests.

tests/test__mongomock.py
```diff
@@ -1704,6 +1704,14 @@ def test__aggregate20(self):
...
+        pipeline = [
+            {'$project': {'_id': 0, 'max': {'$max': [5, 9, 1]},
```
pcorpet: There are no tests including references to values like `$a`, or `$b`. Either add an extra test for it, or cover this case here.

mongomock/collection.py
```diff
@@ -1553,18 +1575,10 @@ def _extend_collection(out_collection, field, expression):
...
-                                        doc_dict[field] = sum(values) / max(len(values), 1)
-                                    elif operator == "$min":
-                                        val_it = (val or MAXSIZE for val in values)
-                                        doc_dict[field] = min(val_it)
-                                    elif operator == "$max":
-                                        val_it = (val or -MAXSIZE for val in values)
-                                        doc_dict[field] = max(val_it)
+                                    if operator in aggregate_operators:
+                                        doc_dict[field] = _handle_aggregate_operator(operator,
+                                                                                     values,
```
pcorpet: I suggest you break after `(` that would make a more readable code.

mongomock/collection.py
```diff
@@ -1553,18 +1575,10 @@ def _extend_collection(out_collection, field, expression):
...
+                                                                                     doc_dict)
                                     elif operator == "$first":
```
pcorpet: I suggest you use the `group_operators` few line above (l.1565 in your new code) and that you have the final else to that chain here handling group operators that are not implemented yet.

mongomock/collection.py
```diff
@@ -1454,13 +1480,9 @@ def _handle_date_operator(operator, values, doc_dict):
...
 
         def _handle_project_operator(operator, values, doc_dict):
-            if operator == '$min':
-                if len(values) > 2:
-                    raise NotImplementedError("Although %d is a valid amount of elements in "
-                                              "aggregation pipeline, it is currently not "
-                                              " implemented in Mongomock" % len(values))
-                return min(_parse_expression(values[0], doc_dict),
-                           _parse_expression(values[1], doc_dict))
+
```
pcorpet: Please drop that blank line.

mongomock/collection.py
```diff
@@ -1359,6 +1367,24 @@ def aggregate(self, pipeline, **kwargs):
             '$millisecond',
             '$dateToString']
 
+        def _handle_aggregate_operator(operator, values, doc_dict):
+            if operator == "$sum":
+                val_it = (_parse_expression(val, doc_dict) or 0 for val in values)
```
pcorpet: Actually, I've realized that you parse those expression twice in the case of a group, which is not a good idea. I'll try to factorize but without this problem.

PR #373, opened 2018-01-18
Avoid useless data copy in collection when only looking for document count

touilleMan:
This provide a pretty serious boost when inserting data (given each insert has to search for the number of document with the id of the newly inserted document)


vmalloc:
@touilleMan thanks!

Issue #378, opened 2018-02-11
Collation not supported for `Collection.find()`


CalgaryMichael:
I have received the following error when attempting to call mongomock `Collection.find()` via pyMODM:
```
Traceback (most recent call last):
  File "/Users/calgarymichael/_git/ubqt/ubqt-core/tests/unit_tests/mock_tests/test_mock_queryset.py", line 34, in test_raw
    self.assertTrue(isinstance(results[0], TestModel))
  File "/Users/calgarymichael/.envs/ubqt/lib/python3.6/site-packages/pymodm/queryset.py", line 563, in __getitem__
    return clone.skip(key).first()
  File "/Users/calgarymichael/.envs/ubqt/lib/python3.6/site-packages/pymodm/queryset.py", line 127, in first
    return next(iter(self.limit(-1)))
  File "/Users/calgarymichael/.envs/ubqt/lib/python3.6/site-packages/pymodm/queryset.py", line 543, in __iter__
    return (to_instance(doc) for doc in self._get_raw_cursor())
  File "/Users/calgarymichael/.envs/ubqt/lib/python3.6/site-packages/pymodm/queryset.py", line 533, in _get_raw_cursor
    collation=self._collation)
TypeError: find() got an unexpected keyword argument 'collation'
```

It seems pretty straightforward: add `collation=None` as a keyword to the `find` method.

A quick look showed a similar issue had been brought up in issue #369. Judging by the MongoDB documentation [here](https://docs.mongodb.com/manual/reference/collation/), Collation seems to be unsupported in mongomock on a variety of the listed functions. It may be beneficial to add full functionality for Collation in a future release.

### Relevant info:
Python: 3.6.1
mongomock: 3.9.0
PyMongo: 3.6.0
pymodm: 0.4.0

PR #379, opened 2018-02-12
added collation as an attribute to the Cursor object

CalgaryMichael:
Relevant issue: #378

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1746,7 +1746,8 @@ def _resolve_sort_key(key, doc):
 
 class Cursor(object):
 
-    def __init__(self, collection, spec=None, sort=None, projection=None, skip=0, limit=0, collation=None):
+    def __init__(self, collection, spec=None, sort=None, projection=None, skip=0, limit=0,
+                 collation=None):
         super(Cursor, self).__init__()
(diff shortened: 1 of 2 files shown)


vmalloc:
@CalgaryMichael  the build is failing...

CalgaryMichael:
Looks like the build is passing

vmalloc:
I don't see any functional difference being implemented here with or without collation... Am I missing something? Please see the note regarding stubs here: https://github.com/mongomock/mongomock/#important-note-about-project-status--development

CalgaryMichael:
This commit does not intend to add any functionality. My intention was to merely fix an incompatibility issue with pyMODM. I did not add the functionality of Collation as I was unsure if that was a direction that was desired.

Essentially what I am doing here is the same thing that was done in [this commit](https://github.com/mongomock/mongomock/commit/447bc8710a6f57f23517ebd53c68004763177055).

vmalloc:
Ok, but I'm still not super happy with this. The commit you mentioned was merged when I was less available as a maintainer, and it still goes a bit against what Mongomock tries to achieve. However I will merge this as it sounds like there shouldn't be much harm from this particular case. Thanks for contributing!

PR #380, opened 2018-02-16
Add support to tz_aware=True option

touilleMan:
Currently only tz naive dates are returned by mongomock, this patch allow to return tz aware datetime

### DIFF
--- a/mongomock/helpers.py
+++ b/mongomock/helpers.py
@@ -7,8 +7,8 @@ import warnings
 
 
 try:
-    from bson import (ObjectId, RE_TYPE)
-    from bson.tz_util import utc
+    from bson import (ObjectId, RE_TYPE, tz_util)
+    utc = tz_util.utc
 except ImportError:
     from mongomock.object_id import ObjectId  # noqa
     RE_TYPE = re._pattern_type
--- a/mongomock/helpers.py
+++ b/mongomock/helpers.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timedelta, tzinfo
 from mongomock import InvalidURI
 import re
 from six.moves.urllib_parse import unquote_plus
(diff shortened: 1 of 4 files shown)


mongomock/helpers.py
```diff
@@ -8,10 +8,34 @@
 
 try:
     from bson import (ObjectId, RE_TYPE)
+    from bson.tz_util import utc
```
vmalloc: This is an anti-pattern. You don't know which of these statements failed with `ImportError`, and the consequences can be unexpected

touilleMan: fixed ^^

mongomock/helpers.py
```diff
@@ -7,8 +7,8 @@
 
 
 try:
-    from bson import (ObjectId, RE_TYPE)
-    from bson.tz_util import utc
+    from bson import (ObjectId, RE_TYPE, tz_util)
```
vmalloc: This still suffers from the same problem - if only RE_TYPE is missing, you will use your own ObjectId....

touilleMan: I'm not sure to understand you... originally the try/except contains only `from bson import (ObjectId, RE_TYPE)`.
So the fact that the custom version of ObjectId will be used if we cannot import RE_TYPE doesn't come from my PR right ? (not pointing any finger here, it's just to understand :smile: )

I would say it's not a big concern given the bson module is pretty stable so the try/except guard is only to fallback if the module is not present, not to allow mixed configuration when only a part of the module is present.

Anyway, if you prefer I can modify the PR to do the try/except per element imported:
```python
try:
    from bson import ObjectId
except ImportError:
    class ObjectId:
        # whatever

try:
    from bson import RE_TYPE
except ImportError:
    RE_TYPE = ...

try:
    from bson.tz_util import utc
except ImportError:
    ...
```

What do you say ? :stuck_out_tongue:

vmalloc: You're right - it wasn't introduced in this PR - so not holding any blame here :smile: 

I would very much like to see the change you included above. I think it will make the code better. Thanks!

touilleMan: I've updated the PR :smiley:

vmalloc:
@touilleMan looking good! will merge soon. Thanks for contributing!

PR #381, opened 2018-02-17
Fix datetime millisecond resolution according to MongoDB capabilities

touilleMan:
MongoDB date resolution is millisecond, but Python datetime object goes up to microsecond.
This PR lower the resolution of the datetime objects stored in the fake collection to mimic this behavior.


vmalloc:
Thanks!

Issue #382, opened 2018-03-25
PyMODM compatibility


CalgaryMichael:
I have been attempting to create an API wrapper to mock out query sets for PyMODM using this library. Unfortunately, there are a few areas of incompatibility between these two libraries that will raise runtime errors. The good news is that most of the issues that I have found are a simple fix of adding `**kwargs` to some function parameters, and one instance of needing to cast a generator to a list.

I would love for this package to be compatible with PyMODM. I have the fixes for all the hiccups that I know of, so creating a pull request for this would be quick. Let me know what you think!

vmalloc:
PRs are very welcome! I try to go through them in a timely manner, but usually I get to them within a few days. Make sure you include unit tests to those changes on separate PRs, and follow the [mocking guideline in the README](https://github.com/mongomock/mongomock#important-note-about-project-status--development)

pcorpet:
@CalgaryMichael, please see cb5eb8c, it should have included all the basics you were after. If there are still incompatibilities, please reopen this bug and add details.

PR #386, opened 2018-04-08
Basic pyMODM compatibility

CalgaryMichael:
This is a simple pull request for #382 which is focused on making mongomock compatible with pyMODM. This should address all immediate compatibility issues and allow for basic functionality with these two libraries.

As I continue to use mongomock and pyMODM together, I am bound to find more bugs. If the maintainers are comfortable with the following plan, I would like to tackle these discrepancies with their own pull requests as they come up.

Before moving forward with this, I believe the question of testing should be addressed. It would be possible to build a test suite within this library that tests mongomock against pyMODM. However, I assume that this is outside the scope and responsibility that this library wants to hold.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -303,11 +303,11 @@ class Collection(object):
         validate_write_concern_params(**kwargs)
         return self._insert(data)
 
-    def insert_one(self, document):
+    def insert_one(self, document, **kwargs):
         validate_is_mutable_mapping('document', document)
         return InsertOneResult(self._insert(document), acknowledged=True)
 
-    def insert_many(self, documents, ordered=True):
+    def insert_many(self, documents, ordered=True, **kwargs):
         if not isinstance(documents, collections.Iterable) or not documents:
             raise TypeError('documents must be a non-empty list')
         for document in documents:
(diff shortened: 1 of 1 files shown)


vmalloc:
This PR goes a bit against the guideline mentioned above mocking [here](https://github.com/mongomock/mongomock#important-note-about-project-status--development). I also mentioned that guideline on the relevant issue back when it was opened...

CalgaryMichael:
I am not entirely sure which section of the guideline that this PR conflicts with. This PR merely attempts to smooth out issues with incompatible API, as these methods are not accepting certain inputs.

In my opinion, it would seem best that a mocking API is capable of accepting any arbitrary input, but only acts upon a selection of inputs for the outcome of the API. That being said, I do not have as full of a picture, so I am sure there is a consideration missing in my opinion.

vmalloc:
The main goal of mongomock is indeed compatibility, but not at the cost of stability or test correctness. I’ll try to demonstrate with a very simple example. 

Let’s say someone is testing his or her code against mongomock. At some point the code introduces a call to insert_many using the `ordered` parameter:

```
db.collection.insert_many(coll, orderd=False)
```
Of course the line above has a typo. That typo will cause the actual code to break in production.  The purpose of mongomock is to catch such errors during testing. Merging this PR will cause the code to pass in testing. 

But the same also goes beyond mere argument names. Features should try to behave like the real thing as much as possible, especially when there are noticeable side effects (let’s say a flag that should modify the collection item). I’m not familiar with MODM in particular though so it might not have such side effects.

pcorpet:
@CalgaryMichael hopefully #345, that was merged recently, covers your use case. Please open an issue if other arguments are missing.

epicfaace:
How do you initialize pymodm with mongomock? What url would you pass in to pymodm.connect (https://pymodm.readthedocs.io/en/latest/getting-started.html#connecting-to-mongodb)

PR #387, opened 2018-04-12
Unit test that {x: {$in: [None,...]}} matches when {x: {$exists: False}}

dwinston:
`{x: {$in: [None,...]}}` should match both when x is `None`, when x is an array with `None` in it, *and* when x is not present. This PR tests for that. I have a separate PR for the fix, as you said should be done in a comment in #382.

### DIFF
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -586,9 +586,11 @@ class MongoClientCollectionTest(_CollectionComparisonTest):
         self.cmp.do.insert([
             dict(x=single),
             dict(x=even),
-            dict(x=prime)])
+            dict(x=prime),
+            dict()])
         self.cmp.compare_ignore_order.find({'x': {'$in': [7, 8]}})
         self.cmp.compare_ignore_order.find({'x': {'$in': [4, 5]}})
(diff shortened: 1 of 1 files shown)


vmalloc:
@dwinston Hi! It seems like the Travis tests fail on this PR. Can you please look into it and see why?

vmalloc:
@dwinston oops, I just noticed there was a second PR related to this one... Sorry about that!

I merged the other PR. Can you rebase this one so that we can make sure it passes?

Thanks!

vmalloc:
Thanks!

dwinston:
Great! What is the procedure for PyPI release, and how can I help? I see that 3.9.0 was released Feb 7, and 3.80 was released Mar 2017.

vmalloc:
Currently there is no real release procedure, and it's based on demand. I can release a version if you want.

dwinston:
That would be great, so that I can use a released version in my code that
uses the feature I added in my PR.

vmalloc:
Released 3.10.0. Should be on pypi soon

dwinston:
Yay, thanks!

PR #388, opened 2018-04-12
Ensure {x: {$in: [None,...}} matches if {x: {$exists: False}}

dwinston:
This PR satisfies the failing test introduced by #387.


vmalloc:
Thanks!

PR #392, opened 2018-04-30
Handle $push for fields nested inside an array

### LINKED ISSUE #391: $push does not work inside another array
When trying to update a document with the $push command, it cannot locate the field to update if the field is inside an array itself:
```py
{'$push': {'messages.0.authors': 'me'}}
```
The code chokes on the "0", with an error:
`TypeError: list indices must be integers or slices, not str`

I'll try sending a fix if I get the time (I've patched it for now in `mongomock/collection.py` line 603):
```py
for field in nested_field_list[:-1]:
    if isinstance(subdocument, dict):
        if field not in subdocument:
            subdocument[field] = {}
        subdocument = subdocument[field]
    else:
        subdocument = subdocument[int(field)]
```

pcorpet:
Fixes #391.

TDD: The tests fail without the fix https://travis-ci.org/mongomock/mongomock/builds/373252736

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -599,9 +599,12 @@ class Collection(object):
                             # create nested attributes if they do not exist
                             subdocument = existing_document
                             for field in nested_field_list[:-1]:
-                                if field not in subdocument:
-                                    subdocument[field] = {}
-                                subdocument = subdocument[field]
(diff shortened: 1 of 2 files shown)


vmalloc:
Thanks for your contribution!

PR #394, opened 2018-06-06
Fix nested field unique index in insert

touilleMan:
Currently nested unique index are not handled properly, to reproduce the error:
```python
cl = mongomock.MongoClient()
col = cl.foo.foo

col.create_index('a.b', unique=True)
col.insert({'a': {'b': 1}})
col.insert({'a': {'b': 1}})  # Doesn't raise DuplicateKeyError
```

btw @vmalloc  could it be possible to get commit rights on the repo ? I'm now pretty confident with the codebase (see [my PR history on the project](https://github.com/mongomock/mongomock/pulls?q=is%3Apr+author%3AtouilleMan+is%3Aclosed))
From my point of view anybody with a couple of PRs merged should be granted with this right, this is a really good way to keep the project alive and from my personal experience, people are always very respectful and willing to follow the guidelines when put in this position.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -379,22 +379,6 @@ class Collection(object):
     def _internalize_dict(self, d):
         return {k: copy.deepcopy(v) for k, v in iteritems(d)}
 
-    def _has_key(self, doc, key):
-        key_parts = key.split('.')
-        sub_doc = doc
-        for part in key_parts:
-            if part not in sub_doc:
-                return False
-            sub_doc = sub_doc[part]
-        return True
-
-    def _remove_key(self, doc, key):
-        key_parts = key.split('.')
-        sub_doc = doc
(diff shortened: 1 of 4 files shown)


vmalloc:
@touilleMan I can make you a collaborator, no problem. I see your contributions so far and I think you're totally up for the task. I only want to turn your attention to the one rule I strictly care about - no mocking without side effects (the rule in the README stating that "empty" implementations are only allowed if they don't have side effects in reality). Let me know if you agree and I'll add you to the project.

touilleMan:
> Let me know if you agree and I'll add you to the project.

It's all good for me, I agree it's much better to be explicit (i.e. crash early with `NotImplementedError`) about the limit of the project

vmalloc:
@touilleMan Cool! I added you as a collaborator on the repository. Thanks a lot!

PR #396, opened 2018-06-19
Fix unique check on update

touilleMan:
Update currently do not run unique indexes checks. This PR fixes this ;-)


vmalloc:
Thanks!

Issue #312, opened 2017-03-16
insert_many does not raise BulkWriteError


omridanan:
Hi,

When using insert_many method, it raises DuplicateKeyError in case of one of the keys already exists in the collection, instead of BulkWriteError.

PR #393, opened 2018-05-28
added collection call options for 3.6 compatibility

gpernot:
Just add bypass_document_validation= and session= options to collection calls for 3.6 compatibility

PR #398, opened 2018-07-05
Fix handling of duplicate key errors when inserting many documents and

pcorpet:
pymongo lib is present.

For #312.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -334,14 +334,25 @@ class Collection(object):
             raise TypeError('documents must be a non-empty list')
         for document in documents:
             validate_is_mutable_mapping('document', document)
-        try:
-            return InsertManyResult(self._insert(documents), acknowledged=True)
-        except DuplicateKeyError:
-            raise BulkWriteError('batch op errors occurred')
+        return InsertManyResult(self._insert(documents), acknowledged=True)
 
     def _insert(self, data):
         if isinstance(data, list) or isinstance(data, types.GeneratorType):
-            return [self._insert(item) for item in data]
+            results = []
+            for index, item in enumerate(data):
+                try:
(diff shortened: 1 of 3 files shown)


mongomock/collection.py
```diff
@@ -277,7 +277,8 @@ def execute(self, write_concern=None):
     def add_insert(self, doc):
         self.insert(doc)
 
-    def add_update(self, selector, doc, multi, upsert, collation=None):
+    def add_update(self, selector, doc, multi, upsert, collation=None,
+                   array_filters=None):
```
pcorpet: This change is not related to my fix, however the tests with pymongo lib present would fail if this wasn't there. I can split it in a different commit/PR if you prefer.

vmalloc: What does `array_filters` do on real pymongo/mongodb?

pcorpet: https://docs.mongodb.com/manual/reference/method/Bulk.find.arrayFilters/
It determines which array elements to modify for an update operation on an array field.

vmalloc: So it's a bit problematic to mock it without doing anything. It means there is some use case that would work differently on mongomock and on the real implementation...

pcorpet: I see. I thought it was OK, as it's exactly what happens with the collation arg that is not used either. I found the discussion in https://github.com/mongomock/mongomock/pull/393 which explains why we don't want to add it.

I'm dropping it.

I still find puzzling that tests do not pass...

vmalloc:
Thanks for contributing!

mongomock/collection.py
vmalloc: Yes, I agree there are a lot of confusing things in how Pymongo affects environments.

PR #399, opened 2018-07-08
Fix bulk updating when not using array filters.

pcorpet:
I've left the exception only if the user is indeed using an array filter.


vmalloc:
Thanks!

PR #402, opened 2018-07-11
Update Missing_Features.rst

ScriptProdigy:
Getting error "NotImplementedError: Although '$multiply' is a valid aritmetic operator for the aggregation pipeline, it is currently not implemented  in Mongomock." when using $multiply in aggregation pipelines in my tests :(


vmalloc:
Thanks!

PR #408, opened 2018-07-31
Fixes $project stage in aggregation

jheden:


tests/test__collection_api.py
```diff
@@ -1532,6 +1532,45 @@ def test__aggregate_project_out(self):
         self.assertEqual(expect, new_actual)
         self.assertEqual(expect, list(old_actual))
 
+    def test__aggregate_project_include_in_exclusion(self):
+        self.db.collection.insert_one({'_id': 1, 'a': 2, 'b': 3})
+        with self.assertRaises(ValueError) as err:
+            self.db.collection.aggregate([
+                {'$match': {'_id': 1}},
+                {'$project': {
+                    'a': False,
```
pcorpet: Note that the order of iteration is not fixed in Python, therefore your two tests might just do the same thing. If you really want to test your code, you can import collections and use an OrderedDict. I'm also fine keeping only one of those two tests, or just leave it like that if you're lazy ;-)

tests/test__collection_api.py
```diff
@@ -1532,6 +1532,45 @@ def test__aggregate_project_out(self):
...
+        actual = self.db.collection.aggregate([
+            {'$match': {'_id': 1}},
+            {'$project': {
+                'a': True,
+                'b': True,
+                '_id': False
+            }}
+        ])
+        self.assertEqual(
+            [{'a': 2, 'b': 3}],
```
pcorpet: This should fit on one line.

mongomock/collection.py
```diff
@@ -1754,15 +1754,40 @@ def _extend_collection(out_collection, field, expression):
...
-                            filter_list.append(field)
+                        if field == '_id':
+                            continue
+                        if method is None:
+                            method = 'include' if value else 'exclude'
+                            if method == 'exclude' and include_id:
+                                raise ValueError(
+                                    "Bad projection specification, cannot include fields "
+                                    "or add computed fields during an exclusion projection: %s" % v)
+                        elif method == 'include' and not value and field != '_id':
```
pcorpet: No need to check for field not being _id as you would already skip the whole loop (see `continue` line 1763).

mongomock/collection.py
```diff
@@ -1754,15 +1754,40 @@ def _extend_collection(out_collection, field, expression):
...
+                        elif method == 'exclude' and value:
+                            raise ValueError(
+                                "Bad projection specification, cannot include fields "
+                                "or add computed fields during an exclusion projection: %s" % v)
+                        filter_list.append(field)
+                        if value:
                             out_collection = _extend_collection(out_collection, field, value)
-                    out_collection = [{k: v for (k, v) in x.items() if k in filter_list}
-                                      for x in out_collection]
+                    if (method == 'include' and include_id is not False) or \
```
pcorpet: Here you seem to assume that `method` is either `include` or `exclude`, however if there's only `_id` in the dict, then it's not defined. Can you fix this please?

pcorpet (changes requested):
Thanks a lot for this fix and the nice tests. I have few comments before we can merge your PR.

mongomock/collection.py
jheden: You are absolutely right. This should now be fixed by line 1760.

mongomock/collection.py
```diff
@@ -1754,15 +1754,41 @@ def _extend_collection(out_collection, field, expression):
                                 unwound_collection[-1], v[1:], field_item)
                     out_collection = unwound_collection
                 elif k == '$project':
-                    filter_list = ['_id']
+                    filter_list = []
+                    method = None
+                    include_id = v.get('_id')
+                    if not v:
+                        method = 'include' if include_id else 'exclude'
```
pcorpet: Actually `$project` should raise `pymongo.errors.OperationFailure: Invalid $project :: caused by :: specification must have at least one field` if `v` is empty.

I think the case that was not handled properly below was the case where `v` only includes `_id`. I think the easiest is to move the check at the bottom of the for loop and check whether `method` is set or not yet.

pcorpet (changes requested):

mongomock/collection.py
jheden: Yeah, sorry. Previously I was using `pop` to grab `_id`, and when I did this change I still had that in mind.

pcorpet (approved):

PR #409, opened 2018-07-31
Adds $lookup stage in aggregation

jheden:
Adds the Equality Match feature described at https://docs.mongodb.com/manual/reference/operator/aggregation/lookup/

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1650,13 +1650,37 @@ class Collection(object):
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
                 elif k == '$lookup':
+                    for operator in ('let', 'pipeline'):
+                        if operator in stage['$lookup']:
+                            raise NotImplementedError(
+                                "Although '%s' is a valid lookup operator for the "
+                                "aggregation pipeline, it is currently not "
+                                "implemented in Mongomock." % operator)
+                    for operator in ('from', 'localField', 'foreignField', 'as'):
+                        if operator not in stage['$lookup']:
+                            raise OperationFailure(
+                                "Must specify '%s' field for a $lookup" % operator)
+                        if not isinstance(stage['$lookup'][operator], str):
+                            raise OperationFailure(
+                                'Arguments to $lookup must be strings')
+                        if operator in ('localField', 'foreignField') and \
+                                stage['$lookup'][operator].startswith('$'):
+                            raise OperationFailure(
+                                "FieldPath field names may not start with '$'")
+                        if operator in ('localField', 'as') and \
+                                '.' in stage['$lookup'][operator]:
+                            raise NotImplementedError(
+                                "Although '.' is valid in the 'localField' and 'as' "
+                                "parameters for the lookup stage of the aggregation "
+                                "pipeline, it is currently not implemented in Mongomock.")
+
                     foreign_name = stage['$lookup']['from']
                     local_field = stage['$lookup']['localField']
                     foreign_field = stage['$lookup']['foreignField']
                     local_name = stage['$lookup']['as']
                     foreign_collection = self.database.get_collection(foreign_name)
                     for doc in out_collection:
-                        query = doc[local_field]
+                        query = doc.get(local_field)
                         if type(doc[local_field]) is list:
                             query = {'$in': query}
                         matches = foreign_collection.find({foreign_field: query})
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1663,7 +1663,7 @@ class Collection(object):
                         if not isinstance(stage['$lookup'][operator], str):
                             raise OperationFailure(
                                 'Arguments to $lookup must be strings')
-                        if operator in ('localField', 'foreignField') and \
+                        if operator in ('as', 'localField', 'foreignField') and \
                                 stage['$lookup'][operator].startswith('$'):
                             raise OperationFailure(
                                 "FieldPath field names may not start with '$'")
@@ -1681,7 +1681,7 @@ class Collection(object):
                     foreign_collection = self.database.get_collection(foreign_name)
                     for doc in out_collection:
                         query = doc.get(local_field)
-                        if type(doc[local_field]) is list:
+                        if isinstance(query, list):
                             query = {'$in': query}
                         matches = foreign_collection.find({foreign_field: query})
                         doc[local_name] = [foreign_doc for foreign_doc in matches]
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1649,6 +1649,18 @@ class Collection(object):
                 if k == '$match':
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
+                elif k == '$lookup':
+                    foreign_name = stage['$lookup']['from']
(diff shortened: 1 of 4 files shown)


mongomock/collection.py
```diff
@@ -1649,6 +1649,18 @@ def _extend_collection(out_collection, field, expression):
                 if k == '$match':
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
+                elif k == '$lookup':
+                    foreign_name = stage['$lookup']['from']
```
pcorpet: If any of those fields is missing, it should raise a `pymongo.errors.OperationFailure: must specify 'from' field for a $lookup` instead of a simple `KeyError`.

mongomock/collection.py
```diff
@@ -1649,6 +1649,18 @@ def _extend_collection(out_collection, field, expression):
...
+                    local_field = stage['$lookup']['localField']
```
pcorpet: Also note that you don't handle the `pipeline` parameter, so make sure to raise a `NotImplementedError` if pipeline is set.

mongomock/collection.py
```diff
@@ -1649,6 +1649,18 @@ def _extend_collection(out_collection, field, expression):
...
+                    foreign_field = stage['$lookup']['foreignField']
+                    local_name = stage['$lookup']['as']
+                    foreign_collection = self.database.get_collection(foreign_name)
+                    for doc in out_collection:
+                        query = doc[local_field]
```
pcorpet: what if `local_field` is missing? Please catch that case.

mongomock/collection.py
```diff
@@ -1649,6 +1649,18 @@ def _extend_collection(out_collection, field, expression):
...
+                        if type(doc[local_field]) is list:
+                            query = {'$in': query}
+                        matches = foreign_collection.find({foreign_field: query})
```
pcorpet: I'm not sure how `$lookup` should work, please add a link to the MongoDB page describing the feature to your PR. Also if some docs have some weird values in the local_field like `{'$exists': True}`.

mongomock/collection.py
```diff
@@ -1649,6 +1649,18 @@ def _extend_collection(out_collection, field, expression):
                 if k == '$match':
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
+                elif k == '$lookup':
```
pcorpet, line 1652: For such a big new feature, we require that you add an end to end test to make sure that `pymongo` behaves like `mongomock`: add something in `test__mongomock.py` please.

pcorpet (changes requested):
Yet another feature! Thanks Jonathan.

mongomock/collection.py
jheden: I added the link with restriction in my original comment, thank you for the suggestion.
As for the weird values (specifically field names prefixed with `$`), see https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names

mongomock/collection.py
jheden: Since the testing strategy with `MultiCollection` only allows for testing against one collection at a time without major overhaul or hacking, I decided to run the `lookup` against the same collection.

mongomock/collection.py
```diff
@@ -1650,13 +1650,25 @@ def _extend_collection(out_collection, field, expression):
                     out_collection = [doc for doc in out_collection
                                       if filter_applies(v, doc)]
                 elif k == '$lookup':
+                    for operator in ['let', 'pipeline']:
```
pcorpet: I personally prefer using tuples rather than list for such short inlined lists.

mongomock/collection.py
```diff
@@ -1650,13 +1650,25 @@ def _extend_collection(out_collection, field, expression):
...
+                        if not isinstance(stage['$lookup'][operator], str):
+                            raise OperationFailure('arguments to $lookup must be strings')
                     foreign_name = stage['$lookup']['from']
                     local_field = stage['$lookup']['localField']
                     foreign_field = stage['$lookup']['foreignField']
                     local_name = stage['$lookup']['as']
                     foreign_collection = self.database.get_collection(foreign_name)
                     for doc in out_collection:
-                        query = doc[local_field]
+                        query = doc[local_field] if local_field in doc else None
```
pcorpet: I think that a simple `doc.get(local_field)` would do the equivalent.

mongomock/collection.py
pcorpet: Sounds good to me. Thanks for adding.

mongomock/collection.py
```diff
@@ -1649,6 +1649,30 @@ def _extend_collection(out_collection, field, expression):
...
+                            raise NotImplementedError(
+                                "Although %s is a valid lookup operator for the "
+                                "aggregation pipeline, it is currently not "
+                                "implemented in Mongomock." % operator)
+                    for operator in ['from', 'localField', 'foreignField', 'as']:
+                        if operator not in stage['$lookup']:
+                            raise OperationFailure(
+                                "must specify '%s' field for a $lookup" % operator)
+                        if not isinstance(stage['$lookup'][operator], str):
+                            raise OperationFailure('arguments to $lookup must be strings')
```
pcorpet: Please add a blank space after this line to clearly show that you're done with the initial checks.

mongomock/collection.py
```diff
@@ -1649,6 +1649,30 @@ def _extend_collection(out_collection, field, expression):
...
+                        if operator in stage['$lookup']:
+                            raise NotImplementedError(
+                                "Although %s is a valid lookup operator for the "
+                                "aggregation pipeline, it is currently not "
+                                "implemented in Mongomock." % operator)
+                    for operator in ['from', 'localField', 'foreignField', 'as']:
+                        if operator not in stage['$lookup']:
+                            raise OperationFailure(
+                                "must specify '%s' field for a $lookup" % operator)
+                        if not isinstance(stage['$lookup'][operator], str):
```
pcorpet: You should also raise an error `pymongo.errors.OperationFailure: FieldPath field names may not start with '$'.` if localField or foreignField starts with a `$`.

mongomock/collection.py
```diff
@@ -1649,6 +1649,30 @@ def _extend_collection(out_collection, field, expression):
...
+                            raise OperationFailure('arguments to $lookup must be strings')
+                    foreign_name = stage['$lookup']['from']
+                    local_field = stage['$lookup']['localField']
+                    foreign_field = stage['$lookup']['foreignField']
+                    local_name = stage['$lookup']['as']
+                    foreign_collection = self.database.get_collection(foreign_name)
+                    for doc in out_collection:
+                        query = doc[local_field] if local_field in doc else None
```
pcorpet: Note that `local_field` could contain a `.`. I suggest that for now you just raise a `NotImplementedError` if you see this so that we can get your code merged. If someone else cares about using `.`, they will send another PR.

Also it's the same for `local_name`.

pcorpet (changes requested):
Thanks for the changes. I found some other corner cases which might behave differently than MongoDB.

Also feel free to add yourself in the contributors in the README.md if you wish.

pcorpet (approved):

PR #411, opened 2018-08-07
Add missing arguments to `Collection.bulk_write()`

### LINKED ISSUE #389: Collection.bulk_write needs an ordered argument
pymongo's bulk_write method on Connection objects takes a second parameter, a boolean called ordered that defaults to True. mongomock should implement this. You don't need to do anything at all with the value of the parameter, just take it.

drewbrew:
Bring the method signature up to par with pymongo's:

- Add `ordered`, `bypass_document_validation`, and `session`
- Rename `operations` to `requests`

Fixes #389.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1788,7 +1788,7 @@ class Collection(object):
     def rename(self, new_name, **kwargs):
         self.database.rename_collection(self.name, new_name, **kwargs)
 
-    def bulk_write(self, operations):
+    def bulk_write(self, operations, ordered=True, bypass_document_validation=False, session=None):
         bulk = BulkOperationBuilder(self)
         for operation in operations:
             operation._add_to_bulk(bulk)
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1788,9 +1788,9 @@ class Collection(object):
     def rename(self, new_name, **kwargs):
         self.database.rename_collection(self.name, new_name, **kwargs)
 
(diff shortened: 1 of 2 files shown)


pcorpet (changes requested):
Hi Drew, thanks for your suggestion.

As documented in our README we ask that features that are not implemented do not keep silent when used: here we would require your code to raise a `NotImplementedError` if any of the new arguments of bulk_write are called without their default values.

drewbrew:
Thanks, @pcorpet. I've added the `NotImplementedError` raises as requested.

mongomock/collection.py
```diff
@@ -1788,9 +1788,15 @@ def with_options(
     def rename(self, new_name, **kwargs):
         self.database.rename_collection(self.name, new_name, **kwargs)
 
-    def bulk_write(self, operations):
+    def bulk_write(self, requests, ordered=True, bypass_document_validation=False, session=None):
+        if not ordered:
+            raise NotImplementedError('Working in unordered mode is unsupported')
```
pcorpet: Please make sure that `mongomock` is the obvious culprint: "Unordered mode is a valid MongoDB operation, however it is not supported by mongomock yet". Same below.

pcorpet (changes requested):

mongomock/collection.py
drewbrew: I added a note about sessions only being valid for MongoDB 3.6+ for maximum clarity as well.

pcorpet (approved):

drewbrew:
Thanks for the merge!

PR #412, opened 2018-08-08
Implement a basic version of write_concern property.

### LINKED ISSUE #406: mongomock not working with mongoengine
Code to reproduce this problem:

```python
from mongoengine import connect, Document, fields

connect('mongoenginetest', host='mongomock://localhost')

class User(Document):
    _id = fields.IntField(primary_key=True)
    name = fields.StringField()

user = User(_id=1, name="Alice")

User.objects.insert(user) # TypeError: 'Collection' object is not callable
```

Looks like it is because mongomock doesn't implement ``write_concerns`` at all. and it calls ``collection.write_concern.document`` while doing insert.

And ``collection.write_concern`` doesn't return a ``WriteConcern`` object, it just returns another collection.

And I think this could solve the problem:

```python
class Collection(...):
    ...
    def __init__(...):
        self.__write_concerns = WriteConcern()

    @property
    def write_concerns(self):
        return self.__write_concerns

# and in write_concerns.py, we can do
class WriteConcern(object):
    __slots__ = ("__document", "__acknowledged", "__server_default")

    def __init__(self, w=None, wtimeout=None, j=None, fsync=None):
        self.__document = {}

    @property
    def document(self):
        return self.__document.copy()

```

Hope my suggestion helps

pcorpet:
Fixes #406.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -376,12 +376,13 @@ class BulkOperationBuilder(object):
 
 class Collection(object):
 
-    def __init__(self, db, name):
+    def __init__(self, database, name, create=False, write_concern=None):
         self.name = name
-        self.full_name = "{0}.{1}".format(db.name, name)
-        self.database = db
+        self.full_name = "{0}.{1}".format(database.name, name)
+        self.database = database
         self._documents = OrderedDict()
-        self._force_created = False
+        self._force_created = create
+        self._write_concern = write_concern or WriteConcern()
         self._uniques = []
 
     def _is_created(self):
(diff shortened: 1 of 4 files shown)


Motor-taxi-master-rider:
I think this commit will solve the problem that: Using mongomock in mongoengine, `update` function will not work well. It is great to have this.

zozoens31:
Please, add a test that WriteConcern() object is immutable (see http://api.mongodb.com/python/current/api/pymongo/write_concern.html#pymongo.write_concern.WriteConcern.document)

pcorpet:
@zozoens31 Done.

zozoens31:
LGTM

vmalloc:
A long shot - but are @zozoens31 and @pcorpet related?

pcorpet:
Not that long :-)
https://www.bayesimpact.org/about
http://fcorpet.free.fr/indexan.html

vmalloc:
Oh I see LOL

Issue #368, opened 2017-12-28
Method doesn`t have validation if document fields is value None or not.


AndrewKhodyakov:
Hello, I use mongomock for my tests - it is very usefull!

But I got some problem with filters gt\ly\dte\lte execution.
If I save document which look like this: `{'some_value': None}` 
and then try to apply filter - `{'some_value':{'gt': 0}}` to this collection - I
get the exception:

> Traceback (most recent call last):
  File "/home/ajet/work_repo/nlp_store/src/tests/flask_tests.py", line 294,
in test_e_therapies_tree_build
    data=json.dumps(tmp))
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/werkzeug/test.py",
line 788, in post
    return self.open(*args, **kw)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/flask/testing.py",
line 127, in open
    follow_redirects=follow_redirects)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/werkzeug/test.py",
line 751, in open
    response = self.run_wsgi_app(environ, buffered=buffered)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/werkzeug/test.py",
line 668, in run_wsgi_app
    rv = run_wsgi_app(self.application, environ, buffered=buffered)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/werkzeug/test.py",
line 871, in run_wsgi_app
    app_rv = app(environ, start_response)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/eve/flaskapp.py",
line 1033, in __call__
    return super(Eve, self).__call__(environ, start_response)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/flask/app.py",
line 1994, in __call__
    return self.wsgi_app(environ, start_response)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/flask/app.py",
line 1985, in wsgi_app
    response = self.handle_exception(e)
  File
"/home/ajet/.virtualenvs/nlp_store/lib/python3.6/site-packages/flask/app.py",
line 1540, in handle_exception
    reraise(exc_type, exc_value, tb)
  File
[...]

PR #415, opened 2018-08-09
Implement Objects comparison and stop trying compare values of differ…

### LINKED ISSUE #400: Api call that uses mongomock causes 500 INTERNAL SERVER ERROR but works with Mongodb
The following code causes an error when using mongomock, but not when using mongodb:

        result = self.logs_engine.LogCollection.objects(
            __raw__={'app_id': app_id, "results": {
        "$gt": [{"$size": "results"}, 0]}}).only(
            "user_input", "app_id", "results", "date_time")
        **print(result)**

The error is: TypeError: '>' not supported between instances of 'dict' and 'dict'

pcorpet:
…ent types. Fixes #400.

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -3,10 +3,15 @@ from datetime import datetime
 from .helpers import ObjectId, RE_TYPE
 from . import OperationFailure
 
+import numbers
 import operator
 import re
 from sentinels import NOTHING
 from six import iteritems, string_types
+try:
+    from types import NoneType
+except ImportError:
+    NoneType = type(None)
 
 COMPILED_RE_TYPE = type(re.compile('a'))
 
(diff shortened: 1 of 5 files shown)


pcorpet:
Also fixes #368.

mongomock/filtering.py
```diff
@@ -167,6 +168,69 @@ def _not_nothing_and(f):
...
+
+    return op(a, b)
+
+
+def _get_compare_type(val):
+    """Get a number representing the base type of the value used for comparison.
+
+    See https://docs.mongodb.com/manual/reference/bson-type-comparison-order/
+    """
+    if isinstance(val, type(None)):
```
vmalloc: Cache this through `from types import NoneType`

vmalloc: Also, doesn't this enumeration exist anywhere??? Perhaps in pymongo?

mongomock/helpers.py
```diff
@@ -269,6 +270,11 @@ def patch_datetime_awareness_in_document(value):
     # mixing tz aware and naive.
     # On top of that, MongoDB date precision is up to millisecond, where Python
     # datetime use microsecond, so we must lower the precision to mimic mongo.
+    if isinstance(value, OrderedDict):
```
vmalloc: The two `if`s can be united by just initializing by the object's type: `type(obj)(...)`

vmalloc (changes requested):

mongomock/filtering.py
pcorpet: Yep, thanks. I didn't find any Python code doing the enumeration. I've linked the mongo C++ implementation and switched to use the same ID (they have the same ordering).

mongomock/helpers.py
pcorpet: Smart. Thanks!

vmalloc (approved):

vmalloc:
LGTM

Issue #349, opened 2017-09-05
Operators inside $sum aggregate operator are not evaluated


florianjourda:
```
mock_db.some_table.aggregate([
        {'$group': {
            '_id': '$someField',
            'countNonEmptyValues': {'$sum': {'$cond': [{'$gt': ['$someOtherField', null]}, 1, 0]}},
        }},
])
```
returns
```
File "/usr/local/lib/python3.6/site-packages/mongomock/collection.py", line 1549, in aggregate
    doc_dict[field] = sum(val_it)
TypeError: unsupported operand type(s) for +: 'int' and 'dict'
```
It seems that `{'$sum': {'$cond': [{'$gt': ['$someOtherField', null]}, 1, 0]}}` is not evaluated (which would give an int) and stays a dict.

pcorpet:
Hi Florian, thanks for reporting. There were several missing features for mongomock to evaluate your expression:
- the first one was indeed the `$sum` not evaluating its argument (this is fixed now)
- a second one which is still there is that `$cond` is not implemented in MongoMock (working on a patch)
- finally the last bit missing is that `$gt` is not good (in mongomock) at comparing values with different types

And of course the error was really not explicit (we try to raise `NotImplementedError` when we know we have blind spots).

Please be patient.

PR #417, opened 2018-08-10
Implement the `$cond` operator in the aggregate pipeline. For #349.

pcorpet:


mongomock/collection.py
```diff
@@ -1581,11 +1581,23 @@ def _handle_conditional_operator(operator, values, doc_dict):
                 except KeyError:
                     return fallback
                 return out_value if out_value is not None else fallback
-            else:
-                raise NotImplementedError(
-                    "Although '%s' is a valid date operator for the "
-                    "aggregation pipeline, it is currently not implemented "
-                    " in Mongomock." % operator)
+            if operator == '$cond':
```
jheden: To be in line with how the other types of operators, and the stages of the aggregation pipeline, are handled, I'd suggest using `elif` here (and maybe keeping the `else` for `NotImplementedError`, even though it is the very last statement in the method).

jheden (approved):
Nicely done!

jheden:
Again, approving despite a failing test, due to the test failing for unrelated circumstances.

mongomock/collection.py
pcorpet: Done.

jheden (approved):

PR #420, opened 2018-08-11
Fix exception in python3 when aggregating with $group on composite _id. Fixes #419

### LINKED ISSUE #419: In python 3 $group (aggregate) with composite _id raises exception
In python 2.7 the following works fine:
```
        collection.insert_many(
            [
                {'a': 2, 'b': 3, 'c': 4},
                {'a': 2, 'b': 3, 'c': 5},
                {'a': 1, 'b': 1, 'c': 1},
            ]
        )
        actual = collection.aggregate([
            {'$group': {'_id': {'a': '$a', 'b': '$b'}}},
        ])
```
In python 3.6 it raises:
```
TypeError: '<' not supported between instances of 'dict' and 'dict'
```

henryykt:


mongomock/collection.py
```diff
@@ -1847,6 +1848,16 @@ def _resolve_sort_key(key, doc):
     return 1, value
 
 
+def _fix_sort_key(key_getter):
+    def fixed_getter(doc):
+        key = key_getter(doc)
+        # Convert dictionaries to make sorted() work in Python 3.
+        if isinstance(key, dict):
+            return [(k, v) for (k, v) in sorted(key.items())]
```
pcorpet, line 1858: In Python 2 `{'a': 1, 'b': 1} > {'a': 2}, so this will change the previous order. Is that intended, was the order broken before?

tests/test__collection_api.py
```diff
@@ -1806,3 +1806,37 @@ def test_aggregate_unwind_push_first(self):
...
+                {'a': 2, 'b': 3, 'c': 4},
+                {'a': 2, 'b': 3, 'c': 5},
+                {'a': 1, 'b': 1, 'c': 1},
+            ]
+        )
+        actual = collection.aggregate([
+            {'$group': {'_id': {'a': '$a', 'b': '$b'}}},
+        ])
+        self.assertEqual(
+            [{'_id': {'a': 1, 'b': 1}}, {'_id': {'a': 2, 'b': 3}}],
```
pcorpet: I agree your code is doing exactly that, however it's not clear that this is how MongoDB would sort it. Especially if one item of the collection is missing a b field. When I test on a real MongoDB, `{'a': 2}` gets sorted before `{'a': 1, 'b': 1}` and that wouldn't be the case with your code. Can you please add a test in `test__mongomock.py` to compare the mongomock behaviour with the one from pymongo + MongoDB please?

pcorpet (changes requested):

tests/test__collection_api.py
henryykt: @pcorpet thanks for taking a look at this PR.

I understand that one of the goals for mongomock is to match MongoDB output as closely as possible, however in the case of the $group operator, the manual states that "$group does not order its output documents."  (see https://docs.mongodb.com/manual/reference/operator/aggregation/group/)

So the order is basically undefined and the user should always use the $sort operator to force order (if order matters).

pcorpet: Indeed, you're right! So the sorting is only useful to use itertools.groupby, it makes total sense.

Can you please add a python comment just above the `out_collection = sorted(out_collection, key=sort_key_getter)` line explaining that (so that the next dev that needs to fix it don't get it wong):

```py
# Sort the collection only for the itertools.groupby. $group does not order its output document.
```

Also maybe add a python comment in the tests to make sure that if someone changes the order and the tests break, they don't panick and just update the test. Bonus points if you can make the tests not enforce the current order (and then drop the comment in the test).

henryykt: Good idea. thx. Updated now.

pcorpet (approved):

PR #422, opened 2018-08-11
 Fix operations that reuse fields existing in the original collection.

pcorpet:


mongomock/aggregate.py
```diff
@@ -622,19 +618,35 @@ def _handle_project_stage(in_collection, unused_database, options):
...
         {
             k: v for (k, v) in x.items()
             if (method == 'include') == (k in filter_list)
         }
-        for x in out_collection
+        for x in in_collection
     ]
+    if new_fields_collection:
+        return [
+            dict(a, **b)
```
zozoens31, line 646: what happens here if there are some common keys between the two dicts? Please, make sure the behaviour is consistent with MongoDB doc, and add test to make sure of it.

pcorpet: It's actually not possible: fields in options can either be added to new_field_collections xor to filter_list (as it's added only once), and in a case of exclusion, new_field_collections is empty so there cannot be a conflict.

Thanks for asking though, I've added a test with an exclusion and a computed field (that raises an error).

PR #423, opened 2018-08-12
Fix the `modified_count` result when update operations do not modify …

### LINKED ISSUE #355: `modified_count` is not correct when an `update` operation does not modify a doc
In real mongodb, when you update an existing document, if the value are exactly the same, then the document will not be updated. and it returns.

```python
col.insert({"_id": 0, "v": 0})

res = col.update({"_id": 0}, {"$set": {"v": 0}})

res["nModified"] # is  0, but in mongomock its 1 !
```

@vmalloc

I want to contribute, but having trouble to run tox test locally, looks like your tox setting is very complicate. Just wants to get more detail about how to run the test, in my understanding:

fork the repo
toggle this forked repo on on travis.ci
commit, and test on travis.ci
no way to test locally using tox
please correct me

pcorpet:
…a document.

Fixes #355.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -561,12 +561,13 @@ class Collection(object):
         updated_existing = False
         upserted_id = None
         num_updated = 0
+        num_matched = 0
         for existing_document in itertools.chain(self._iter_documents(spec), [None]):
             # we need was_insert for the setOnInsert update operation
             was_insert = False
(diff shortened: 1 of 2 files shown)


armudgal:
Looks like travis has some connection problems.

pcorpet:
@kira0204 yep, feel free to review the code anyway and ignore the timeout failure.

PR #427, opened 2018-08-15
Update documentation

pcorpet:
- stop using deprecated methods (insert, update)
 - document how to use mongomock in combination with mock.


armudgal:
LGTM

PR #429, opened 2018-08-17
_inc_updater now handles incrementing values in an array as opposed to dict

dvetal-cigna:
- Allows for proper incrementation of an individual index value in a sub-document array.
- Also updated allowed types in the mongo to mongomocker comparer.  This is because pymongo will return a list with None values when an index to be incremented does not exist in the current mongo array.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1117,13 +1117,7 @@ class Collection(object):
             else:
                 return
         field_name = field_name_parts[-1]
-        if isinstance(doc, list):
-            try:
-                doc[int(field_name)] = field_value
-            except IndexError:
-                pass
-        else:
-            updater(doc, field_name, field_value)
+        updater(doc, field_name, field_value)
 
     def _iter_documents(self, filter=None):
         return (document for document in list(itervalues(self._documents))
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -2079,6 +2080,13 @@ def _set_updater(doc, field_name, value):
             # bson validation
             BSON.encode({field_name: value}, check_keys=True)
         doc[field_name] = value
+    if isinstance(doc, list):
+        try:
+            doc[int(field_name)] = value
+        except IndexError:
+            len_diff = int(field_name) - (len(doc)-1)
```
pcorpet: E226 missing whitespace around arithmetic operator, see [Travis report](https://travis-ci.org/mongomock/mongomock/jobs/417438511) for all lint errors

mongomock/collection.py
```diff
@@ -2079,6 +2080,13 @@ def _set_updater(doc, field_name, value):
             # bson validation
             BSON.encode({field_name: value}, check_keys=True)
         doc[field_name] = value
+    if isinstance(doc, list):
+        try:
+            doc[int(field_name)] = value
```
pcorpet: I suggest you do the `int(field_name)` only once:
```py
field_index = int(field_name)
```

pcorpet: What if field index is negative? I think that it's not correct for the Mongo syntax to have a field `foo.-3.bar` so you don't need to handle it correctly, however mongomock should raise an error instead of handling it silently.

pcorpet: Same for the $inc updater.

mongomock/collection.py
```diff
@@ -1045,7 +1046,7 @@ def _update_document_single_field(self, doc, field_name, field_value, updater):
         field_name = field_name_parts[-1]
         if isinstance(doc, list):
             try:
-                doc[int(field_name)] = field_value
+                updater(doc, field_name, field_value)
```
pcorpet: I'd rather not have this line copied from below. I suggest you handle the case where one silence the`IndexError` in the updater functions.

Edit: after reading your code below, it seems it's already handled and that you won't have any IndexError raised, so no need to differentiate the cases.

pcorpet (changes requested):
Thanks for your contribution.

dvetal-cigna:
OK. I'll take care of this in the next day or two.

mongomock/collection.py
dvetal-cigna: ok

dvetal-cigna:
@pcorpet requested changes made.

pcorpet (approved):

Issue #431, opened 2018-08-20
Mongomock doesn't support count_documents


boxmansf:
See documentation:
http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.count_documents

PR #432, opened 2018-08-21
Implement the new method count_documents introduced in MongoDB v3.7.

pcorpet:
Implements #431.


vmalloc:
@pcorpet  Looks good to me

PR #436, opened 2018-08-25
Add session arg to database and collection methods.

### LINKED ISSUE #435: TypeError: list_indexes() got an unexpected keyword argument 'session'
Hi, currently `collection.list_indexes()` does not expect a `session` parameter due to which this `TypeError` is thrown, could you please check it.
Thanks

pcorpet:
Note that this does not implement anything and will raise
`NotImplementedError` if the new arg is actually used.

Fixes #435.


zozoens31:
LGTM

PR #437, opened 2018-08-26
handling multiple fields inside "only" operation using the same prefix.

sagydr:
E.g running Book.objects().only('book_data.name', 'book_data.length') doesn't work
before this fix because the key_parts[-2] (AKA book_data) is common for those 2 keys so
they override each other.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -806,7 +806,6 @@ class Collection(object):
         return Cursor(self, spec, sort, projection, skip, limit, collation=collation)
 
     def _get_dataset(self, spec, sort, fields, as_class):
-        print "sagy - _get_dataset = %r , %r , %r" % (spec, fields, as_class)
         dataset = (self._copy_only_fields(document, fields, as_class)
                    for document in self._iter_documents(spec))
         if sort:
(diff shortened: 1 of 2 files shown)


pcorpet:
Hi @sagydr , thanks for reporting the error and proposing a fix.

It's not easy to follow as you are using a syntax that is not the one used by `mongomock`. I believe this is `mongoengine` although I'm not familiar with it. I think that in MongoDB equivalent it would be:

```py
db.book.find({}, {
  'book_data.name': 1,
  'book_data.length': 1,
})
```

Can you confirm and add a test please? Note that the test should fail without your fix.

sagydr:
Hi @pcorpet, you are right, my example was mongoengine one. 
i'll fix the syntax and add a test.

sagydr:
Hi @pcorpet, I just pushed a test for my fix, and I see I have conflicts with code which does "_combine_projection_spec" i guess you already merged a fix to the same bug?

pcorpet:
Indeed, I finalized #331 which probably fixes your bug: maybe you can try your new test without changing anything, it should work with the code of the develop branch. I'll try to do a release this week if you confirm that it works for you.

pcorpet:
Oh, and sorry about that, I should have seen the link between the two PR and come back here.

pcorpet:
I confirm that your test now passes with recent versions of mongomock. Closing this PR.

PR #440, opened 2018-08-29
Fix some unicode issues in Python 2.

### LINKED ISSUE #395: Regression with $group operator unicode issue
#222 fixed an unicode issue with $group operator but I recently got a similar issue.

In `_parse_basic_expression` function, replacing `str` with `text_type` resolves it but there is other occurrences that may be concerned.

pcorpet:
Fixes #395


zozoens31:
LGTM

PR #443, opened 2018-09-01
Implement options of the $unwind aggregation operation.

### LINKED ISSUE #354: aggregate $unwind does not handle dict parameter
I am trying to run tests using mockmongo around my dao function which is running an aggregate command (below)

```py
collection1.aggregate([
            {
                "$unwind": {
                    "path": "$collection2",
                    "preserveNullAndEmptyArrays": True
                }
            },
            {
                "$lookup": {
                    "from": "collection2",
                    "localField": "collection2",
                    "foreignField": "_id",
                    "as": "collection2"
                }
            },
            {
                "$match": {"_id": id, "deleted": False}
            }
        ])
```

But the test fails with
'**ValueError: $unwind failed: exception: field path references must be prefixed with a '$' '{'path': '$randomcollection', 'preserveNullAndEmptyArrays': True}**'

Any help would be greatly appreciated

I am using 
mockmongo = 3.8.0
pymongo=3.4.0
mongo=3.4.9

pcorpet:
Fixes #354.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -134,6 +134,23 @@ def set_value_by_dot(doc, key, value):
     return doc
 
 
+def delete_value_by_dot(doc, key):
+    """Delete dictionary value using dotted key"""
+    try:
+        parent_key, child_key = key.rsplit('.', 1)
+        parent = get_value_by_dot(doc, parent_key)
+    except ValueError:
+        child_key = key
+        parent = doc
+
+    if isinstance(parent, dict):
+        del parent[child_key]
+    else:
+        raise KeyError()
+
+    return doc
+
+
 class BulkWriteOperation(object):
     def __init__(self, builder, selector, is_upsert=False):
         self.builder = builder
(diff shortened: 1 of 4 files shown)


zozoens31:
This also changes the behavior of $unwind on non-array field, to comply with https://docs.mongodb.com/manual/reference/operator/aggregation/unwind/#non-array-field-path

mongomock/collection.py
```diff
@@ -1883,26 +1900,44 @@ def _extend_collection(out_collection, field, expression):
...
-                                '"%s", value found: %s' % (v, array_value))
-                        for field_item in array_value:
-                            unwound_collection.append(copy.deepcopy(doc))
-                            unwound_collection[-1] = set_value_by_dot(
-                                unwound_collection[-1], v[1:], field_item)
+                        if isinstance(array_value, list):
+                            iter_array = enumerate(array_value)
+                        else:
+                            iter_array = [(None, array_value)]
+                        if not array_value and should_preserve_null_and_empty:
```
zozoens31: please replace `array_value` by `iter_array`, and add a test to make sure a field with `false` value is preserved.

pcorpet: Done. Although iter_array is never falsy.

PR #444, opened 2018-09-02
Ensure uniqueness of new unique index among documents already existin…

### LINKED ISSUE #351: Creating a unique index for collection doesn't check uniqueness in existing documents
``Collection.create_index`` method doesn't check uniqueness in existing document. 

For example:

if we do: ``col.insert([dict(name="Alice"), dict(name="Alice")])``

then do: ``col.create_index([("name", pymongo.ASCENDING)])``, it suppose to raise error in pymongo, but not raises in mongomock.

Plus, the implementation of unique index can be improved by maintaining a set variable, instead of running a find query every times when insert a new document, it's too slow.

@vmalloc

I want to contribute, but having trouble to run tox test locally, looks like your tox setting is very complicate. Just wants to get more detail about how to run the test, in my understanding:

1. fork the repo
2. toggle this forked repo on on travis.ci
3. commit, and test on travis.ci
4. no way to test locally using tox

please correct me

pcorpet:
…g upon creation.

Fixes #351

### DIFF
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -1015,6 +1015,16 @@ class CollectionAPITest(TestCase):
 
         self.assertEqual(self.db.collection.find({}).count(), 1)
 
+    def test__create_uniq_idxs_with_dupes_already_there(self):
+        self.db.collection.insert({"value": 1})
+        self.db.collection.insert({"value": 1})
+
+        with self.assertRaises(mongomock.DuplicateKeyError):
+            self.db.collection.create_index([("value", 1)], unique=True)
+
+        self.db.collection.insert({"value": 1})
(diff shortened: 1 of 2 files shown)


tests/test__collection_api.py
```diff
@@ -1014,6 +1014,17 @@ def test_sparse_unique_index_dup(self):
             self.db.collection.insert({'value': 'a'})
 
         self.assertEqual(self.db.collection.find({}).count(), 1)
+        self.assertEqual(self.db.collection.find({}).count(), 1)
```
zozoens31: Please drop duplicate line

pcorpet: Done.

PR #445, opened 2018-09-03
Implement comparison of values of different types when using $gt in t…

### LINKED ISSUE #349: Operators inside $sum aggregate operator are not evaluated

pcorpet:
…he aggregation pipeline.

Fixes #349.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -41,8 +41,10 @@ from six import text_type
 
 from mongomock.command_cursor import CommandCursor
 from mongomock import ConfigurationError, DuplicateKeyError, BulkWriteError
+from mongomock.filtering import bson_compare
 from mongomock.filtering import filter_applies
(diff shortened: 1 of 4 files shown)


zozoens31:
LGTM

PR #452, opened 2018-09-27
Introduce a patch method.

pcorpet:
For an easy replacement of pymongo by mongomock across several clients and servers.

Note that I didn't want to add any requirement for the lib, so that it's only available on python 3 or on python 2 if the mock library is also installed.

### DIFF
--- a/mongomock/patch.py
+++ b/mongomock/patch.py
@@ -1,0 +1,132 @@
+from .mongo_client import MongoClient
+from mongomock import InvalidURI
+import time
+
+try:
+    from unittest import mock
+    _IMPORT_MOCK_ERROR = None
+except ImportError:
+    try:
+        import mock
+        _IMPORT_MOCK_ERROR = None
+    except ImportError as error:
+        _IMPORT_MOCK_ERROR = error
+
+try:
+    import pymongo
+    _IMPORT_PYMONGO_ERROR = None
+except ImportError as error:
+    _IMPORT_PYMONGO_ERROR = error
+
+
+def patch(servers='localhost', on_new='error'):
+    """Patch pymongo.MongoClient.
+
+    This will patch the class MongoClient and use mongomock to mock MongoDB
+    servers. It keeps a consistant state of servers across multiple clients so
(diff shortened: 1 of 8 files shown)


mongomock/patch.py
```diff
@@ -0,0 +1,127 @@
...
+        _IMPORT_MOCK_ERROR = error
+
+try:
+    import pymongo
+    _IMPORT_PYMONGO_ERROR = None
+except ImportError as error:
+    _IMPORT_PYMONGO_ERROR = error
+
+
+def patch(on_new='error', servers='localhost'):
```
zozoens31: Please, put `servers` as first parameter (or even as *args), since this is what we patch, on_new being only an option

zozoens31: Maybe use `pymongo` as default `on_new`, so that the patch influences only the described servers

mongomock/patch.py
```diff
@@ -0,0 +1,127 @@
...
+        if on_new == 'pymongo':
+            return PyMongoClient(*args, **kwargs)
+
+        raise ValueError(
+            'MongoDB server %s:%d does not exist.\n' % client.address + '%s' % parsed_servers)
+
+    return mock.patch('pymongo.MongoClient', _create_persistent_client)
+
+
+def _parse_host_and_port(uri, default_port=27017):
```
zozoens31: please, make this one work with replica sets mongouri (or add TODO)

zozoens31 (changes requested):

mongomock/patch.py
pcorpet: Nice. I like it.

pcorpet: I think it's pretty rare to use `mongomock` and wanting to access a real MongoDB server in the same tests. For the usage I've seen, it's more common to want to hear about any unplanned access to Mongo.

mongomock/patch.py
pcorpet: Good call. I've supported them partially and added a TODO to do it fully.

PR #454, opened 2018-10-01
Implements list_collection_names

moonso:
- Adds a warning when using the old `collection_names`
- Implements list_collection_names
- Updates tests

### DIFF
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -39,6 +39,8 @@ class Database(object):
 
     def collection_names(self, include_system_collections=True, session=None):
         warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
+        if include_system_collections:
+            return list(self._get_created_collections())
 
         return self.list_collection_names(session=session)
 
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -39,7 +39,7 @@ class Database(object):
 
     def collection_names(self, include_system_collections=True, session=None):
         warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
-        
+
         return self.list_collection_names(session=session)
 
     def list_collection_names(self, session=None):
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -1,3 +1,5 @@
+import warnings
+
 from collections import OrderedDict
 
 from . import CollectionInvalid
@@ -36,10 +38,13 @@ class Database(object):
         return [name for name, col in self._collections.items() if col._is_created()]
 
     def collection_names(self, include_system_collections=True, session=None):
+        warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
+        
+        return list_collection_names(session=session)
+
+    def list_collection_names(self, session=None):
         if session:
             raise NotImplementedError('Mongomock does not handle sessions yet')
-        if include_system_collections:
-            return list(self._get_created_collections())
 
         result = []
         for name in self._get_created_collections():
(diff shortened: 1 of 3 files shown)


mongomock/database.py
```diff
@@ -36,6 +38,7 @@ def _get_created_collections(self):
         return [name for name, col in self._collections.items() if col._is_created()]
 
     def collection_names(self, include_system_collections=True, session=None):
+        warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
         if session:
```
pcorpet: I suggest you just call
```py
return self.list_collection_names(session=session)
```
instead of duplicating the code.

tests/test__collection_api.py
```diff
@@ -56,12 +56,10 @@ def test__get_collection_full_name(self):
     def test__get_collection_names(self):
```
pcorpet: Please keep at least one test calling collection_names: even if it's deprecated, it's still implemented by pymongo.

pcorpet (changes requested):

tests/test__collection_api.py
moonso: That should be fixed now. Thanks for the input

tests/test__collection_api.py
```diff
@@ -56,12 +56,18 @@ def test__get_collection_full_name(self):
...
         self.db.create_collection('a')
         self.db.create_collection('b')
-        self.assertEqual(set(self.db.collection_names()), set(['a', 'b']))
-        self.assertEqual(set(self.db.collection_names(True)), set(['a', 'b']))
-        self.assertEqual(set(self.db.collection_names(False)), set(['a', 'b']))
+        self.assertEqual(set(self.db.get_collection_names()), set(['a', 'b']))
```
pcorpet: Actually it's collection_names not get_collection_names

pcorpet (changes requested):
Please also rebase and squash all your work in only one commit.

mongomock/database.py
```diff
@@ -36,10 +38,13 @@ def _get_created_collections(self):
...
 
     def collection_names(self, include_system_collections=True, session=None):
+        warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
+
+        return self.list_collection_names(session=session)
+
+    def list_collection_names(self, session=None):
         if session:
             raise NotImplementedError('Mongomock does not handle sessions yet')
-        if include_system_collections:
```
pcorpet: Actually, removing this would be a regression for current users of `mongomock`. I suggest we keep it (for collection_names only) until the collection_names function gets retired.

pcorpet (changes requested):

mongomock/database.py
moonso: Ok so should I let the function work just like in `develop`?

mongomock/database.py
```diff
@@ -36,11 +38,16 @@ def _get_created_collections(self):
         return [name for name, col in self._collections.items() if col._is_created()]
 
     def collection_names(self, include_system_collections=True, session=None):
-        if session:
-            raise NotImplementedError('Mongomock does not handle sessions yet')
+        warnings.warn('collection_names is deprecated. Use list_collection_names instead.')
         if include_system_collections:
```
moonso, line 42: Like this?

pcorpet: Exactly, thanks.

pcorpet (approved):

PR #462, opened 2018-10-26
Implement the $slice operator for $push update operations.

### LINKED ISSUE #257: Add support for $slice modifier in $push update operations
Hi,

I've noticed that the `$slice` operator does not work with mongomock.
If same test if run with a real mongo database this is not a problem.
I can provide with code if necessary...

pcorpet:
Fixes #257.


moonso (approved):
👍

PR #471, opened 2018-11-20
Implement the ping command for database.

### LINKED ISSUE #469: Implement Database.command, at least the ping command
My application health check is sending a ping command but I noticed command method is not implemented in Database, leading to command being interpreted as a Collection.

Maybe we could foresee a first implementation only handling 'ping' command and returning {'ok': 1.0} ?

pcorpet:
Fixes #469.

### DIFF
--- a/tests/test__database_api.py
+++ b/tests/test__database_api.py
@@ -18,7 +18,11 @@ class DatabaseAPITest(TestCase):
         database['_users'].insert_one({'a': 1})
         self.assertEqual(1, database['_users'].find_one().get('a'))
 
+    def test__command_ping(self):
+        database = mongomock.MongoClient().somedb
+        self.assertEqual({'ok': 1}, database.command({'ping': 1}))
+
     def test__command(self):
         database = mongomock.MongoClient().somedb
         with self.assertRaises(NotImplementedError):
-            database.command({'ping': 1})
+            database.command({'count': 'user'})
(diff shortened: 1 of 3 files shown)


Colin-b (approved):
I tried both 'ping' as a string and {'ping': 1} as a dict, both command should work. This implementation is great, clever use of the in keyword to match both string and dict key commands :+1:

tests/test__database_api.py
```diff
@@ -18,7 +18,11 @@ def test__get_collection_by_attribute_underscore(self):
         database['_users'].insert_one({'a': 1})
         self.assertEqual(1, database['_users'].find_one().get('a'))
 
+    def test__command_ping(self):
```
Colin-b, line 21: Just a suggestion, you could maybe add an additional test case to ensure that command as a string works:
```suggestion
    def test__command_ping_as_str(self):
        database = mongomock.MongoClient().somedb
        self.assertEqual({'ok': 1}, database.command('ping'))

    def test__command_ping(self):
```

pcorpet: Woops, you're right. I've also added a test for a string containing "ping" but not "ping".

pcorpet:
:-) It was not on purpose. I had not plan for using it with a string and my implementation would also ping on command strings containing "ping" but not being "ping". Fixed.

PR #474, opened 2018-11-24
Implement with_options when changing write_concern.

### LINKED ISSUE #472: mongomock does not support with_options(write_concern=...)
See below syntax:
write_concern = WriteConcern(w=2, wtimeout = 60000)
colc.with_options(write_concern=write_concern).update_one(...)

I would suggest implementing by just ignoring this param when passed to with_options, as mongomock is not simulating a cluster.

pcorpet:
Fixes #472.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -342,13 +342,9 @@ class BulkOperationBuilder(object):
 
 class Collection(object):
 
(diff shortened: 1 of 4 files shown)


boxmansf (approved):
@pcorpet LGTM. Thank you for fixing it !

PR #478, opened 2018-12-08
Implement the {'$type': 'timestamp'} option of the $currentDate updat…

### LINKED ISSUE #375: Add Support for timestamps in $currentDate
When I update document with {"$currentDate": {"last_update_time": {"$type": "timestamp"}}}

I got a exception: NotImplementedError: timestamp is not supported so far.

Would you have plan to implement this feature? 

Thanks.

pcorpet:
…e operator.

Fixes #375.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -15,9 +15,9 @@ import time
 import warnings
 
 try:
-    from bson import json_util, SON, BSON
+    from bson import json_util, SON, BSON, Timestamp
 except ImportError:
-    json_utils = SON = BSON = None
+    json_utils = SON = BSON = Timestamp = None
 try:
     import execjs
 except ImportError:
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -1801,9 +1796,23 @@ def _min_updater(doc, field_name, value):
         doc[field_name] = min(doc.get(field_name, value), value)
 
 
+_LAST_TIMESTAMP_INC = collections.defaultdict(int)
```
zozoens31: Only the last unix timestamp is needed, not all the used ones (unless we expect going back in time by mocking the `time.time` function).
What do you think about keeping the last used timestamp instead? This might perform better, especially if it is used in an environment with many such updates.

zozoens31 (changes requested):

mongomock/collection.py
pcorpet: Done.

PR #485, opened 2019-01-06
Add a hint method to the Cursor class.

### LINKED ISSUE #205: No 'hint' method for Cursor class
The cursor class has no `hint` method.
Though hinting is not important when using mongomock, this is a problem when trying to use mongomock as a drop-in replacement for pymongo.

http://api.mongodb.org/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.hint

pcorpet:
Fix #205.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1699,6 +1699,14 @@ class Cursor(object):
     def close(self):
         pass
 
+    def hint(self, unused_hint):
+        if self._emitted:
+            raise InvalidOperation('cannot set options after executing query')
+        # TODO(pascal): Once we implement $text indexes and queries, raise an
+        # exception if hint is used on a $text query.
+        # https://docs.mongodb.com/manual/reference/method/cursor.hint/#behavior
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -1699,6 +1699,11 @@ def batch_size(self, count):
     def close(self):
         pass
 
+    def hint(self, unused_hint):
+        if self._emitted:
+            raise InvalidOperation('cannot set options after executing query')
+        return self
```
zozoens31, line 1708: Please also ensure the following
> You cannot use hint() if the query includes a $text query expression.
https://docs.mongodb.com/manual/reference/method/cursor.hint/#behavior

zozoens31 (changes requested):
LGTM

mongomock/collection.py
pcorpet: I've added a TODO because that would require to implement text indexes.

PR #489, opened 2019-01-21
Support read preference arguments (has no effect, but check types).

### LINKED ISSUE #487: Please implement read_preference
Currently this happens when attempting to set read_preference: NotImplementedError: read_preference is a valid parameter for with_options but it is currently not implemented in Mongomock

In my mind this would be a special case where a no-op could be implemented (this is a no-op when running against a single Mongo server as well).

pcorpet:
Closes #487

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -36,7 +36,7 @@ except ImportError:
         BEFORE = False
         AFTER = True
 
-    _READ_PREFERENCE_PRIMARY = None
+    from mongomock.read_preferences import PRIMARY as _READ_PREFERENCE_PRIMARY
 
 from sentinels import NOTHING
 from six import iteritems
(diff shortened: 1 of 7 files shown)


maria-jabon:
> Closes #487

Hi @pcorpet .   Is this intended to ensure when you initiate a client using mongomock that use pass a mongomock.read_preferences.PRIMARY?  Per your commit message this should have no effect but it breaks all my unit tests.

If I instantiate a mongo client using  a pymongo.ReadPreference.PRIMARY I get the following error (which makes no sense because I AM passing a  pymongo.read_preference):


TypeError: read_preference must be an instance of pymongo.read_preference.ReadPreference


This should be allowed (currently throwing the above error):
`
client = MongoClient(
                    mongo_uri, read_preference=pymongo.ReadPreference.PRIMARY_PREFERRED, **kwargs)
`

 I have to change it to this to work:

`
client = MongoClient(
                    mongo_uri, read_preference=mongomock.read_preferences.PRIMARY, **kwargs)
`

pcorpet:
@maria-jabon Thanks for reporting, that's a bad bug. Sorry about it. I'll try to fix it and release a patch ASAP.

pcorpet:
@maria-jabon Please check which version of `pymongo` you are using please.

pcorpet:
I cannot reproduce the error. @maria-jabon can you open a new issue with the code that is failing for you and the versions of the library you are using please?

maria-jabon:
So sorry for the delay.  I will open an issue with the versions and code

PR #494, opened 2019-01-26
Implement the $sort and $position modifiers for the $push update operator.

### LINKED ISSUE #492: The $sort modifier of the $push operator is not implemented
It should be possible to push and sort (and also slice, which would happen after the sort) in one operation. Sort seems to be simply ignored when trying to do so. Test case:

    def test__update_push_and_sort(self):
        self.db.collection.insert_one({'a':
                                       {'b': [{'value': 3},
                                              {'value': 1},
                                              {'value': 2}]}})
        self.db.collection.update_one({}, {'$push': {'a.b': {'$each': [{'value': 4}],
                                                             "$sort": {'value': 1}}}})
        self.assertEqual({u'b': [{u'value': 1}, {u'value': 2}, {u'value': 3}, {u'value': 4}]},
                         self.db.collection.find_one()['a'])

pcorpet:
Fixes #492.

### DIFF
--- a/Missing_Features.rst
+++ b/Missing_Features.rst
@@ -8,7 +8,6 @@ If I miss to include a feature in the below list, Please feel free to add to the
 * create_collection options - https://docs.mongodb.com/v3.2/reference/method/db.createCollection/#definition
 * bypass_document_validation options
 * session options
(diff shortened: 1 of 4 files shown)


ghost:
This fixes the test case failure I was seeing, thanks!

PR #497, opened 2019-01-30
fix DuplicateKeyError when parse None to sparse unique index

Friday21:
I got DuplicateKeyError when I create model with None value for a key which index is unque and sparse. but, with mongoengine, it works fine.
```
class Device(Documnet):
     key_1 = StringField(max_length=256)
     key_2 = StringField(max_length=256)
     
     meta = {
        "collection": "devices",
        "indexes": [
            {
                "fields": ["key_1"],
                "unique": True,
                "sparse": True
            },

d1 = Device(key1='a', key2='b').save()
d2 = Device(key2='b').save()  # key1 is None
d3 = Device(key2='c').save()  # key1 is None
```

I got DuplicateKeyError when I create d3 use mongomock

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1328,6 +1328,8 @@ class Collection(object):
                         if is_sparse:
                             continue
                         index.append(None)
+                if is_sparse and not index:
+                    continue
                 index = tuple(index)
                 if index in indexed:
                     raise DuplicateKeyError('E11000 Duplicate Key Error', 11000)
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -131,6 +131,17 @@ class DatabaseGettingTest(TestCase):
         result = collection.find({})
         self.assertEqual(result.count(), 3)
 
+    def test__sparse_unique_index(self):
+        db = self.client.somedb
+        collection = db.a
+        collection.create_index([('value', 1)], unique=True, sparse=True)
+
+        collection.insert({'value': 'should_be_unique'})
+        collection.insert({'simple': 'simple_without_value'})
+        collection.insert({'simple': 'simple_without_value2'})
+
+        collection.ensure_index([('value', 1)], unique=True, sparse=True)
+
     def test__alive(self):
         self.assertTrue(self.client.alive())
 
--- a/README.rst
+++ b/README.rst
@@ -190,6 +190,7 @@ Also, many thanks go to the following people for helping out, contributing pull 
 * tipok
 * waskew (waskew _at_ narrativescience.com)
 * jmsantorum (jmsantorum [at] gmail [dot] com)
+* lidongyong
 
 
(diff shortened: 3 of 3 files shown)


pcorpet:
Thanks for reporting and already suggesting a fix.

Did you check that your test was failing without the modification?
I've significantly modified the code since version 3.14 and when I tried your test at `HEAD` of the `develop` branch, I do not get the `DuplicateKeyError`.

I'll try to release v3.15 sometime this week but in the meantime, you can check if your `mongoengine` test passes with the latest code by installing it:
```sh
pip install git+https://github.com/mongomock/mongomock@develop
```

Friday21:
> Thanks for reporting and already suggesting a fix.
> 
> Did you check that your test was failing without the modification?
> I've significantly modified the code since version 3.14 and when I tried your test at `HEAD` of the `develop` branch, I do not get the `DuplicateKeyError`.
> 
> I'll try to release v3.15 sometime this week but in the meantime, you can check if your `mongoengine` test passes with the latest code by installing it:
> 
> ```shell
> pip install git+https://github.com/mongomock/mongomock@develop
> ```

My bad, I remove one line code in test which I thought is irrelevant, but it turns out it matters the test result.
```
collection.ensure_index([('value', 1)], unique=True, sparse=True)
```
I will fix the commit

pcorpet:
Got it. It fails indeed.

mongomock/collection.py
```diff
@@ -1331,7 +1331,8 @@ def create_index(self, key_or_list, cache_for=300, session=None, **kwargs):
                 index = tuple(index)
                 if index in indexed:
                     raise DuplicateKeyError('E11000 Duplicate Key Error', 11000)
-                indexed.add(index)
+                if index:
```
pcorpet: The reason this works is not obvious here. I suggest you make it explicit:
```py
if is_sparse and not index:
  continue
```
Note that you can move that above line 1331, as there is no need to convert to a tuple or check if it's in indexed either.

tests/test__mongomock.py
```diff
@@ -131,6 +131,18 @@ def test__drop_database_indexes(self):
         result = collection.find({})
         self.assertEqual(result.count(), 3)
 
+    def test__sparse_unique_index(self):
+        db = self.client.somedb
+        collection = db.a
+        collection.create_index('simple')
```
pcorpet: Please drop that line, the test works without it, and it could lead to confusion.

pcorpet (approved):
Thanks again Friday. Feel free to also add yourself to the contributors in `README.rst`.

mongomock/collection.py
Friday21: yeah, your way is better

tests/test__mongomock.py
Friday21: ok

Friday21:
> Thanks again Friday. Feel free to also add yourself to the contributors in `README.rst`.

OK, love to.

PR #501, opened 2019-02-19
Support max_time_ms in find()

### LINKED ISSUE #500: Support for max_time_ms in find()
I'm getting `find() got an unexpected keyword argument 'max_time_ms'` and I think adding `max_time_ms=None` to `def find()` would solve it easily as a first step, even without an actual implementation yet.

orik123:
Fixes #500

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -912,9 +912,9 @@ class Collection(object):
         spec = filter
         if spec is None:
             spec = {}
-
         validate_is_mapping('filter', spec)
-        return Cursor(self, spec, sort, projection, skip, limit, collation=collation).max_time_ms(max_time_ms)
+        return Cursor(self, spec, sort, projection, skip, limit,
+                      collation=collation).max_time_ms(max_time_ms)
 
     def _get_dataset(self, spec, sort, fields, as_class):
         dataset = self._iter_documents(spec)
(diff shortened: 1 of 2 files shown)


pcorpet:
Thanks for the suggestion.

Please add:
* [ ] a test that would fail before your change
* [ ] a check for the type of `max_time_ms` if it's not `None`
* [ ] a comment on why we don't need an implementation (probably along the lines of the one in `Cursor.max_time_ms` method.

mongomock/collection.py
```diff
@@ -907,12 +907,14 @@ def _discard_operators(self, doc):
...
-             batch_size=0, manipulate=True, collation=None, session=None):
+             batch_size=0, manipulate=True, collation=None, session=None,
+             max_time_ms=None):
         spec = filter
         if spec is None:
             spec = {}
         validate_is_mapping('filter', spec)
-        return Cursor(self, spec, sort, projection, skip, limit, collation=collation)
+        return Cursor(self, spec, sort, projection, skip, limit,
+                      collation=collation).max_time_ms(max_time_ms)
```
pcorpet, line 917: Smart :-)

pcorpet (approved):
Thanks for the changes.

orik123:
Sure, my pleasure! Any idea when will be the next release?

pcorpet:
Erf, I just did one yesterday…
I'll wait at least a week to see if there are bugs in this one and then publish a small fix with your change.

orik123:
Great, thank you!

orik123:
Thank you!

PR #506, opened 2019-03-08
Put logical operator expansion back in a loop

### LINKED ISSUE #505: Problem with $or in 3.15.0 but not 3.14.0
```
$ python2.7
Python 2.7.16 (default, Mar  4 2019, 09:01:38)
[GCC 4.2.1 Compatible Apple LLVM 10.0.0 (clang-1000.11.45.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import mongomock
>>> db = mongomock.MongoClient().db
>>> db.test.insert_one({'x': 1, 'y': 1, 'z': 1})
<pymongo.results.InsertOneResult object at 0x1090b5368>
>>> [x for x in db.test.find()]
[{'y': 1, 'x': 1, 'z': 1, '_id': ObjectId('5c82a21324fb1674d1ebeb96')}]
>>> [x for x in db.test.find({'$or': [{'x': 1}, {'y': 2}], 'z': 2})]
[{'y': 1, 'x': 1, 'z': 1, '_id': ObjectId('5c82a21324fb1674d1ebeb96')}]
>>> [x for x in db.test.find({'$or': [{'x': 2}, {'y': 2}], 'z': 2})]
[]
```
After the `$or` condition my` 'z': 2` condition is ignored but only since 3.15.0
After reverting to 3.14.0 with `pip install mongomock==3.14.0` I get my expected results:
```
>>> import mongomock
>>> db = mongomock.MongoClient().db
>>> db.test.insert_one({'x': 1, 'y': 1, 'z': 1})
<pymongo.results.InsertOneResult object at 0x10e917fc8>
>>> [x for x in db.test.find()]
[{'y': 1, 'x': 1, 'z': 1, '_id': ObjectId('5c82a2bf24fb167549c94d7a')}]
>>> [x for x in db.test.find({'$or': [{'x': 1}, {'y': 2}], 'z': 2})]
[]
>>> [x for x in db.test.find({'$or': [{'x': 2}, {'y': 2}], 'z': 2})]
[]
```
Also reproducible with python3.7

finnhughes:
Fixes #505

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -29,7 +29,13 @@ def filter_applies(search_filter, document):
         # Top level operators.
         if key == '$comment':
             continue
-        if key.startswith('$') and key not in LOGICAL_OPERATOR_MAP:
+        if key in LOGICAL_OPERATOR_MAP:
+            if not search:
+                raise OperationFailure('BadValue $and/$or/$nor must be a nonempty array')
+            if not LOGICAL_OPERATOR_MAP[key](document, search):
+                return False
+            continue
+        if key.startswith('$'):
             raise OperationFailure('unknown top level operator: ' + key)
 
         is_match = False
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -29,11 +29,7 @@ def filter_applies(search_filter, document):
         # Top level operators.
         if key == '$comment':
             continue
-        if key in LOGICAL_OPERATOR_MAP:
-            if not search:
-                raise OperationFailure('BadValue $and/$or/$nor must be a nonempty array')
-            return LOGICAL_OPERATOR_MAP[key](document, search)
-        if key.startswith('$'):
+        if key.startswith('$') and key not in LOGICAL_OPERATOR_MAP:
             raise OperationFailure('unknown top level operator: ' + key)
 
         is_match = False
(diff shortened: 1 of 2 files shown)


finnhughes:
Quick fix, quick check, your welcome to reject and do something that suits you better.

Note travis failure appears to be pylint in collections which this commit doesn't touch

tests/test__collection_api.py
```diff
@@ -2170,6 +2170,14 @@ def test_find_with_comment(self):
         actual = list(self.db.collection.find({'_id': 1, '$comment': 'test'}))
         self.assertEqual([{'_id': 1}], actual)
 
+    def test__find_or_and(self):
+        self.db.collection.insert_many([
+            {'x': 1, 'y': 1},
+        ])
+        self.assertEqual(
+            [],
+            [d['x'] for d in self.db.collection.find({'y': 2, '$or': [{'x': 1}, {'x': 2}]})])
```
pcorpet: Your test only fails randomly as the key of a dict are not sorted. I've replaced your search filter by an OrderedDict that forces the `$or` key to come first.

mongomock/filtering.py
```diff
@@ -61,6 +57,10 @@ def filter_applies(search_filter, document):
                 ) and search) or doc_val == search
             elif isinstance(search, RE_TYPE) and isinstance(doc_val, (string_types, list)):
                 is_match = _regex(doc_val, search)
+            elif key in LOGICAL_OPERATOR_MAP:
```
pcorpet, line 66: It's not sound to check for the `key` in a loop for which it's constant, so I've pushed a fix that only slightly modified the previous behavior (do not return True if it matches, just check the next key).

pcorpet (commented):
Hi @finnhughes and thanks a lot for both reporting and proposing a fix. I've tweaked your solution but hopefully still fix your usecase.

finnhughes:
Thanks, I can confirm all my tests passing with your fix 👍

PR #517, opened 2019-04-17
Added string operations support

ymoran00:
Added support for aggregation operators - all string operations that weren't supported yet, and a new $toString operation that was added in mongodb 4.0

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -5,7 +5,6 @@ import collections
 import copy
 import datetime
 import itertools
-
 import math
 import numbers
 import random
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -6,6 +6,8 @@ import copy
 import datetime
 import functools
 import itertools
+import warnings
+
 import math
 import numbers
 import random
@@ -277,6 +279,9 @@ class _Parser(object):
             length = self.parse(values[2])
             if string is None or first < 0:
                 return ''
+            if length < 0:
+                warnings.warn('Negative length given to $substr is accepted only until '
+                              'MongoDB 3.7. This behavior will vhange in the future.')
             second = len(string) if length < 0 else first + length
             return string[first:second]
         if operator == '$strcasecmp':
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -5,11 +5,12 @@ import collections
 import copy
 import datetime
 import itertools
-import warnings
 
 import math
 import numbers
 import random
+import warnings
+
 import six
 from six import moves
 
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -277,11 +277,15 @@ class _Parser(object):
             string = str(self.parse(values[0]))
             first = self.parse(values[1])
             length = self.parse(values[2])
-            if string is None or first < 0:
+            if string is None:
+                return ''
+            if first < 0:
+                warnings.warn('Negative starting point given to $substr is accepted only until '
+                              'MongoDB 3.7. This behavior will change in the future.')
                 return ''
             if length < 0:
                 warnings.warn('Negative length given to $substr is accepted only until '
-                              'MongoDB 3.7. This behavior will vhange in the future.')
+                              'MongoDB 3.7. This behavior will change in the future.')
             second = len(string) if length < 0 else first + length
             return string[first:second]
         if operator == '$strcasecmp':
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -86,8 +86,6 @@ string_operators = [
     '$substr',
     '$toLower',
     '$toUpper',
-]
-misc_operators = [
     '$toString',
 ]
 comparison_operators = [
@@ -162,8 +160,6 @@ class _Parser(object):
                 return self._handle_set_operator(k, v)
             if k in string_operators:
                 return self._handle_string_operator(k, v)
-            if k in misc_operators:
-                return self._handle_misc_operator(k, v)
             if k in boolean_operators + \
                     text_search_operators + projection_operators:
                 raise NotImplementedError(
@@ -263,24 +259,18 @@ class _Parser(object):
             'aggregation pipeline, it is currently not implemented '
             ' in Mongomock.' % operator)
 
-    def _handle_misc_operator(self, operator, values):
-        if operator == '$toString':
-            parsed = self.parse(values)
-            return str(parsed) if parsed is not None else None
-            # This should never happen: it is only a safe fallback if something went wrong.
-        raise NotImplementedError(  # pragma: no cover
-            "Although '%s' is a valid operator for the aggregation "
-            'pipeline, it is currently not implemented  in Mongomock.' % operator)
-
     def _handle_string_operator(self, operator, values):
         if operator == '$toLower':
-            return str(self.parse(values)).lower()
+            parsed = self.parse(values)
+            return str(parsed).lower() if parsed is not None else ''
         if operator == '$toUpper':
-            return str(self.parse(values)).upper()
+            parsed = self.parse(values)
+            return str(parsed).upper() if parsed is not None else ''
         if operator == '$concat':
             return ''.join([str(self.parse(value)) for value in values])
         if operator == '$substr':
-            assert len(values) == 3, 'substr must have 3 items'
+            if len(values) != 3:
+                raise OperationFailure('substr must have 3 items')
             string = str(self.parse(values[0]))
             first = self.parse(values[1])
             length = self.parse(values[2])
@@ -289,9 +279,13 @@ class _Parser(object):
             second = len(string) if length < 0 else first + length
             return string[first:second]
         if operator == '$strcasecmp':
-            assert len(values) == 2, 'strcasecmp must have 2 items'
+            if len(values) != 2:
+                raise OperationFailure('strcasecmp must have 2 items')
             a, b = str(self.parse(values[0])), str(self.parse(values[1]))
             return 0 if a == b else -1 if a < b else 1
+        if operator == '$toString':
+            parsed = self.parse(values)
+            return str(parsed) if parsed is not None else None
         # This should never happen: it is only a safe fallback if something went wrong.
         raise NotImplementedError(  # pragma: no cover
             "Although '%s' is a valid string operator for the aggregation "
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -86,7 +86,7 @@ string_operators = [
     '$substr',
     '$toLower',
     '$toUpper',
-    '$toString',
+    # '$toString',
 ]
 comparison_operators = [
     '$cmp',
@@ -267,7 +267,8 @@ class _Parser(object):
             parsed = self.parse(values)
             return str(parsed).upper() if parsed is not None else ''
         if operator == '$concat':
-            return ''.join([str(self.parse(value)) for value in values])
+            parsed_list = [self.parse(value) for value in values]
+            return None if None in parsed_list else ''.join([str(x) for x in parsed_list])
         if operator == '$substr':
             if len(values) != 3:
                 raise OperationFailure('substr must have 3 items')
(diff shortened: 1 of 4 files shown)


mongomock/aggregate.py
```diff
@@ -87,6 +87,9 @@
     '$toLower',
     '$toUpper',
 ]
+misc_operators = [
+    '$toString',
```
pcorpet, line 90: Let's add it to the string operators (the string operations convert non-strings to string already).

mongomock/aggregate.py
```diff
@@ -157,7 +160,11 @@ def parse(self, expression):
                 return self._handle_conditional_operator(k, v)
             if k in set_operators:
                 return self._handle_set_operator(k, v)
-            if k in boolean_operators + string_operators + \
+            if k in string_operators:
+                return self._handle_string_operator(k, v)
+            if k in misc_operators:
+                return self._handle_misc_operator(k, v)
+            if k in boolean_operators + \
```
pcorpet, line 164: Please update the Missing_features.rst file to mention that those features are not missing anymore.

tests/test__collection_api.py
```diff
@@ -2597,6 +2597,36 @@ def test__aggregate_arithmetic(self):
             [{'sum': 20.5, 'prod': 30, 'trunc': 1}],
             [{k: v for k, v in doc.items() if k != '_id'} for doc in actual])
 
+    def test__aggregate_string_operations(self):
+        self.db.collection.insert_one({
+            'a': 'Hello',
+            'b': 'World',
+        })
+        actual = self.db.collection.aggregate([{'$project': {
```
pcorpet, line 2616: Please also add that test to test__mongomock.py to make sure the behavior is consistant with Pymongo.

tests/test__collection_api.py
```diff
@@ -2597,6 +2597,36 @@ def test__aggregate_arithmetic(self):
...
+            'sub2': {'$substr': ['$a', -1, 3]},
+            'lower': {'$toLower': '$a'},
+            'upper': {'$toUpper': '$a'},
+        }}])
+        self.assertEqual(
+            [{'concat': 'Hello Dear World', 'sub1': 'Hell', 'sub2': '', 'lower': 'hello', 'upper': 'HELLO'}],
+            [{k: v for k, v in doc.items() if k != '_id'} for doc in actual])
+
+    def test__aggregate_misc_operations(self):
+        _id = ObjectId()
```
pcorpet: Please use an ObjectId for which you choose the hash:
```py
_id = ObjectId('5abcfad1fbc93d00080cfe66')

...

assertEqual([{'toString1: '5abcfad1fbc93d00080cfe66', 'toString2': '3'}], ...)
```

pcorpet: The rationale is that I prefer to keep the logic in tests to their minimum.

tests/test__collection_api.py
pcorpet: Same for the test below.

mongomock/aggregate.py
```diff
@@ -256,6 +263,40 @@ def _handle_comparison_operator(self, operator, values):
...
+
+    def _handle_string_operator(self, operator, values):
+        if operator == '$toLower':
+            return str(self.parse(values)).lower()
+        if operator == '$toUpper':
+            return str(self.parse(values)).upper()
+        if operator == '$concat':
+            return ''.join([str(self.parse(value)) for value in values])
+        if operator == '$substr':
+            assert len(values) == 3, 'substr must have 3 items'
```
pcorpet: Please be consistant with the error that is raised by pymongo. It usually raises OperationFailure instead of assertion errors.

Same below.

mongomock/aggregate.py
```diff
@@ -256,6 +263,40 @@ def _handle_comparison_operator(self, operator, values):
...
+            string = str(self.parse(values[0]))
+            first = self.parse(values[1])
+            length = self.parse(values[2])
+            if string is None or first < 0:
+                return ''
+            second = len(string) if length < 0 else first + length
```
pcorpet, line 288: Please add a test case for negative length.

mongomock/aggregate.py
```diff
@@ -256,6 +263,40 @@ def _handle_comparison_operator(self, operator, values):
...
+            parsed = self.parse(values)
+            return str(parsed) if parsed is not None else None
+            # This should never happen: it is only a safe fallback if something went wrong.
+        raise NotImplementedError(  # pragma: no cover
+            "Although '%s' is a valid operator for the aggregation "
+            'pipeline, it is currently not implemented  in Mongomock.' % operator)
+
+    def _handle_string_operator(self, operator, values):
+        if operator == '$toLower':
+            return str(self.parse(values)).lower()
```
pcorpet: This would fail if values returns None, please add a test for that.

pcorpet (changes requested):
Thanks Yoav, very much appreciated. I have a few comments to keep your code to the existing standards.

tests/test__collection_api.py
ymoran00: the $toString is only supported since mongodb 4.0. Should it be there as well anyway?

mongomock/aggregate.py
ymoran00: np

mongomock/aggregate.py
ymoran00: np

tests/test__collection_api.py
ymoran00: Added for now - unless you'll say otherwise.

tests/test__collection_api.py
ymoran00: good idea. np

mongomock/aggregate.py
ymoran00: np

mongomock/aggregate.py
ymoran00: Apparently according to MongoDB doc - it should return '' for None. So I updated the code and added a test accordingly.

mongomock/aggregate.py
ymoran00: forgot it on push - so added the test and pushed again.

mongomock/aggregate.py
```diff
@@ -256,6 +259,38 @@ def _handle_comparison_operator(self, operator, values):
...
 
+    def _handle_string_operator(self, operator, values):
+        if operator == '$toLower':
+            parsed = self.parse(values)
+            return str(parsed).lower() if parsed is not None else ''
+        if operator == '$toUpper':
+            parsed = self.parse(values)
+            return str(parsed).upper() if parsed is not None else ''
+        if operator == '$concat':
+            return ''.join([str(self.parse(value)) for value in values])
```
pcorpet: What if one of those subvalue is None? would it crash? Please add a test for this.

tests/test__collection_api.py
pcorpet: Oh, you're right, Travis does not have MongoDB 4 yet. Please add a TODO there to reenabe once Travis runs Mongo v4+.

Still add the test for the string operators.

mongomock/aggregate.py
ymoran00: You're right - that's another case. According to docs this will return a concat result of None. Fixed and added test.

tests/test__collection_api.py
ymoran00: np

mongomock/aggregate.py
```diff
@@ -256,6 +259,40 @@ def _handle_comparison_operator(self, operator, values):
...
+            return str(parsed).lower() if parsed is not None else ''
+        if operator == '$toUpper':
+            parsed = self.parse(values)
+            return str(parsed).upper() if parsed is not None else ''
+        if operator == '$concat':
+            parsed_list = [self.parse(value) for value in values]
+            return None if None in parsed_list else ''.join([str(x) for x in parsed_list])
+        if operator == '$substr':
+            if len(values) != 3:
+                raise OperationFailure('substr must have 3 items')
```
pcorpet, line 275: Please add a test for this.

mongomock/aggregate.py
```diff
@@ -256,6 +259,40 @@ def _handle_comparison_operator(self, operator, values):
...
+            first = self.parse(values[1])
+            length = self.parse(values[2])
+            if string is None or first < 0:
+                return ''
+            second = len(string) if length < 0 else first + length
+            return string[first:second]
+        if operator == '$strcasecmp':
+            if len(values) != 2:
+                raise OperationFailure('strcasecmp must have 2 items')
+            a, b = str(self.parse(values[0])), str(self.parse(values[1]))
```
pcorpet, line 293: Add two test for stcasecmp, one for this, and one for when it succeeds and one for when it fails.

mongomock/aggregate.py
```diff
@@ -256,6 +259,40 @@ def _handle_comparison_operator(self, operator, values):
...
+            return 0 if a == b else -1 if a < b else 1
+        # TODO(ymoran00): Re-enable once TRAVIS supports MongoDB 4
```
pcorpet: Sorry, I meant that you should add this TODO on the test in test__mongomock: the feature is welcome and can be tested in test__collection_api

tests/test__collection_api.py
```diff
@@ -2597,6 +2597,41 @@ def test__aggregate_arithmetic(self):
...
+            'lower_err': {'$toLower': None},
+            'upper': {'$toUpper': '$a'},
+            'upper_err': {'$toUpper': None},
+        }}])
+        self.assertEqual(
+            [{'concat': 'Hello Dear World', 'concat_none': None, 'sub1': 'Hell', 'sub2': '',
+              'sub3': 'llo', 'lower': 'hello', 'lower_err': '', 'upper': 'HELLO', 'upper_err': ''}],
+            [{k: v for k, v in doc.items() if k != '_id'} for doc in actual])
+
+    # TODO(ymoran00): Re-enable once TRAVIS supports MongoDB 4
```
pcorpet: Keep this for now.

ymoran00: ok

mongomock/aggregate.py
ymoran00: ok

mongomock/aggregate.py
ymoran00: ok

mongomock/aggregate.py
ymoran00: ok

ymoran00:
@pcorpet Mongo 3.7 behaves differently with negative value in $substr. Before 3.7 negative values were accepted, in 3.7 they cause errors. What should I implement then?

pcorpet:
Let's implement the old behavior for now (to match the version of Mongo used by Travis), and add a warning when it's called that this behavior is changing.

tests/test__mongomock.py
```diff
@@ -2275,7 +2275,25 @@ def test__aggregate33(self):
             'maxNone': {'$max': [None, None]},
             'minNone': {'$min': [None, None]},
             'avgNone': {'$avg': ['a', None]},
-            'sumNone': {'$sum': ['a', None]},
+            'sumNone': {'$sum': ['a', None]}
```
pcorpet: Nit - let's keep the trailing comma here please.

pcorpet:
Thanks for all the changes. Please make sure the CI passes before I merge this.

ymoran00:
> Let's implement the old behavior for now (to match the version of Mongo used by Travis), and add a warning when it's called that this behavior is changing.

Unit testing the old behavior in test__mongomock causes an error when the tox tests mongo 3.7... Maybe I'll remove unit testing for negative values for substr?

pcorpet:
Yes, the test__mongomock is not really a unit test (it actually starts a MongoDB server in the back and compares the behaviors). Keep the unit tests in test__collection_api to be similar to the new behavior and do not test it in test__mongomock at all.

mongomock/aggregate.py
```diff
@@ -5,6 +5,8 @@
 import copy
 import datetime
 import itertools
+import warnings
```
pcorpet: Please move after the other std library imports (after random but before six), and move the six imports (third party) as a different import group.

pcorpet (approved):

PR #523, opened 2019-04-30
A cleaner pull request to add $graphlookup

ymoran00:


mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+    if not isinstance(options.get('maxDepth', 0), six.integer_types):
+        raise OperationFailure(
+            "Argument 'maxDepth' to $graphLookup must be a number")
+    if not isinstance(options.get('restrictSearchWithMatch', {}), dict):
+        raise OperationFailure(
+            "Argument 'restrictSearchWithMatch' to $graphLookup must be a Dictionary")
+    if not isinstance(options.get('depthField', ''), six.string_types):
+        raise OperationFailure(
+            "Argument 'depthField' to $graphlookup must be a string")
+    for operator in ('from', 'startWith', 'connectFromField', 'connectToField', 'as'):
```
pcorpet: please sort the operators alphabetically in your list.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+                "Must specify '%s' field for a $graphLookup" % operator)
+        if not isinstance(options[operator], six.string_types):
+            raise OperationFailure(
+                "Argument '%s' to $graphLookup must be string" % operator)
+        if operator in ('as', 'connectFromField', 'connectToField') and \
+                options[operator].startswith('$'):
+            raise OperationFailure(
+                "FieldPath field names may not start with '$'")
+        if operator == 'startWith' and \
+                not options[operator].startswith('$'):
```
pcorpet: This should fit on the line above.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+        if operator not in options:
+            raise OperationFailure(
+                "Must specify '%s' field for a $graphLookup" % operator)
+        if not isinstance(options[operator], six.string_types):
+            raise OperationFailure(
+                "Argument '%s' to $graphLookup must be string" % operator)
+        if operator in ('as', 'connectFromField', 'connectToField') and \
+                options[operator].startswith('$'):
+            raise OperationFailure(
+                "FieldPath field names may not start with '$'")
```
pcorpet: This should fit on the line above.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+        if operator == 'startWith' and \
+                not options[operator].startswith('$'):
+            raise OperationFailure(
+                "Field ref names must start with '$'")
```
pcorpet: This should fit on the line above.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+            "Argument 'restrictSearchWithMatch' to $graphLookup must be a Dictionary")
+    if not isinstance(options.get('depthField', ''), six.string_types):
+        raise OperationFailure(
+            "Argument 'depthField' to $graphlookup must be a string")
+    for operator in ('from', 'startWith', 'connectFromField', 'connectToField', 'as'):
+        if operator not in options:
+            raise OperationFailure(
+                "Must specify '%s' field for a $graphLookup" % operator)
+        if not isinstance(options[operator], six.string_types):
+            raise OperationFailure(
```
pcorpet, line 529: Actually startWith could be an array, according to the doc. Please add a NotImplementedError for that case.

mongomock/aggregate.py
```diff
@@ -809,13 +892,13 @@ def _handle_count_stage(in_collection, database, options):
...
     '$facet': None,
     '$geoNear': None,
-    '$graphLookup': None,
     '$group': _handle_group_stage,
     '$indexStats': None,
     '$limit': lambda c, d, o: c[:o],
     '$listLocalSessions': None,
     '$listSessions': None,
     '$lookup': _handle_lookup_stage,
+    '$graphLookup': _handle_graph_lookup_stage,
```
pcorpet: please keep it alphabetically sorted.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+    connect_from_field = options['connectFromField']
+    connect_to_field = options['connectToField']
+    local_name = options['as']
+    max_depth = options.get('maxDepth', None)
+    depth_field = options.get('depthField', None)
+    restrict_search_with_match = options.get('restrictSearchWithMatch', {})
+    foreign_collection = database.get_collection(foreign_name)
+    for doc in in_collection:
+        found_items = set()
+        if start_with in doc:
```
pcorpet: Actually start_with is an expression, so it could be `{'$concat`: ['$field1', '$field2']}`, please use parse_expression. That way you can also accept dots in startWith.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+            query = doc.get(start_with)
+            if isinstance(query, list):
+                query = {'$in': query}
+            matches = foreign_collection.find({connect_to_field: query})
+            # TODO(pascal): Do not modify the input collection.
```
pcorpet: Alternatively you could just copy the dict quickly by creating an out_doc, with `out_doc = dict(doc)`

pcorpet: Note, if you do so, please add a test to make sure the initial collection is not modified.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+                        new_match[depth_field] = 0
+                    newly_discovered_matches.append(new_match)
+                    found_items.add(new_match['_id'])
+            doc[local_name] = newly_discovered_matches
+        else:
+            doc[local_name] = []
+        origin_matches = doc[local_name]
+        depth = 1
+        while origin_matches and (max_depth is None or depth <= max_depth):
+            newly_discovered_matches = []
```
pcorpet, line 572: Don't repeat yourself, please create a sub function.

mongomock/aggregate.py
```diff
@@ -509,6 +509,89 @@ def _handle_lookup_stage(in_collection, database, options):
...
+            for match in origin_matches:
+                query = match.get(connect_from_field)
+                if isinstance(query, list):
+                    query = {'$in': query}
+                # if filter:
```
pcorpet: "Do or do not, there is no try." Please drop the commented code.

pcorpet (changes requested):

mongomock/aggregate.py
ymoran00: You're right - good catch.

ymoran00: Added support for any expression in startWith.

mongomock/aggregate.py
ymoran00: Done that after your previous comment on startWith being array and such :)

mongomock/aggregate.py
ymoran00: Added the dict() part, but any idea how to unit test that?

ymoran00:
I've done refactoring according to your suggestions, and added support to startWith expressions.

ymoran00:
did you take a look at the current code?

mongomock/aggregate.py
```diff
@@ -498,15 +498,86 @@ def _handle_lookup_stage(in_collection, database, options):
     foreign_field = options['foreignField']
     local_name = options['as']
     foreign_collection = database.get_collection(foreign_name)
-    for doc in in_collection:
+    out_doc = copy.deepcopy(in_collection)
```
pcorpet: No need to deep copy, we are only modifying the first level, and a deep copy can be quite long. Please leave out from that PR, I'll take care of it for the lookup stage.

mongomock/aggregate.py
```diff
@@ -498,15 +498,86 @@ def _handle_lookup_stage(in_collection, database, options):
...
+    connect_from_field = options['connectFromField']
+    connect_to_field = options['connectToField']
+    local_name = options['as']
+    max_depth = options.get('maxDepth', None)
+    depth_field = options.get('depthField', None)
+    restrict_search_with_match = options.get('restrictSearchWithMatch', {})
+    foreign_collection = database.get_collection(foreign_name)
+    out_doc = copy.deepcopy(in_collection)
+
+    def find_matches_for_depth(query):
```
pcorpet: Let's make it private (prefix by _) here

mongomock/aggregate.py
```diff
@@ -498,15 +498,86 @@ def _handle_lookup_stage(in_collection, database, options):
...
+        found_items = set()
+        depth = 0
+        result = _parse_expression(start_with, doc)
+        newly_discovered_matches = find_matches_for_depth(result)
+        origin_matches = doc[local_name] = newly_discovered_matches
+        while origin_matches and (max_depth is None or depth < max_depth):
+            depth += 1
+            newly_discovered_matches = []
+            for match in origin_matches:
+                result = match.get(connect_from_field)
```
pcorpet: Please do not reuse the same var name for different things (result line 569 is still defined in this scope).

mongomock/aggregate.py
```diff
@@ -498,15 +498,86 @@ def _handle_lookup_stage(in_collection, database, options):
...
+    foreign_name = options['from']
+    start_with = options['startWith']
+    connect_from_field = options['connectFromField']
+    connect_to_field = options['connectToField']
+    local_name = options['as']
+    max_depth = options.get('maxDepth', None)
+    depth_field = options.get('depthField', None)
+    restrict_search_with_match = options.get('restrictSearchWithMatch', {})
+    foreign_collection = database.get_collection(foreign_name)
+    out_doc = copy.deepcopy(in_collection)
```
pcorpet: As I mentioned above, we might be able to avid a deepcopy: we need to copy two levels only, the list of doc and the first level. As this PR has been opened for a while, I suggest you keep a TODO for me to speed up the copy.

mongomock/aggregate.py
```diff
@@ -498,15 +498,86 @@ def _handle_lookup_stage(in_collection, database, options):
...
+    def find_matches_for_depth(query):
+        if isinstance(query, list):
+            query = {'$in': query}
+        matches = foreign_collection.find({connect_to_field: query})
+        new_matches = []
+        for new_match in matches:
+            if filtering.filter_applies(restrict_search_with_match, new_match) \
+                    and new_match['_id'] not in found_items:
+                if depth_field is not None:
+                    new_match[depth_field] = depth
```
pcorpet: I suspect that this would actually modify the foreign collection. Please use:
`new_match = collections.OrderedDict(new_match, **{depth_field: depth})`

pcorpet (changes requested):

pcorpet (approved):

PR #524, opened 2019-05-01
Supporting $replaceRoot operation

ymoran00:


Missing_Features.rst
```diff
@@ -21,12 +21,8 @@ If I miss to include a feature in the below list, Please feel free to add to the
   * `$indexStats <https://docs.mongodb.com/manual/reference/operator/aggregation/indexStats/>`_
   * `$listLocalSessions <https://docs.mongodb.com/manual/reference/operator/aggregation/listLocalSessions/>`_
   * `$listSessions <https://docs.mongodb.com/manual/reference/operator/aggregation/listSessions/>`_
-  * `$replaceRoot <https://docs.mongodb.com/manual/reference/operator/aggregation/replaceRoot/>`_
   * `$sortByCount <https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/>`_
 * Operators within the aggregate pipeline:
-  * Arithmetic operations on dates:
```
pcorpet: Please keep those (not related to your PR).

tests/test__collection_api.py
```diff
@@ -2180,6 +2180,69 @@ def test__find_or_and(self):
...
+            {'_id': 2, 'pets': {'hamsters': 3, 'cats': 4}}
+        ])
+        with self.assertRaises(mongomock.OperationFailure) as err:
+            self.db.a.aggregate([
+                {'$replaceRoot': {
+                    'newRoot': '$not_here'
+                }}
+            ])
+        self.assertEqual(
+            'Could not complete $replaceRoot with given expression.',
```
pcorpet: This is not the error I get when I run your example against a real Mongo server. I get:
```
pymongo.errors.OperationFailure: 'newRoot' expression must evaluate to an object, but resulting value was: MISSING. Type of resulting value: 'missing'. Input document: {}
```

tests/test__collection_api.py
```diff
@@ -2180,6 +2180,69 @@ def test__find_or_and(self):
...
+        ])
+        actual = self.db.a.aggregate([
+            {'$replaceRoot': {'newRoot': '$pets'}}
+        ])
+        self.assertListEqual([
+            {'dogs': 2, 'cats': 3},
+            {'hamsters': 3, 'cats': 4}
+        ], list(actual))
+
+    def test__aggregate_replace_root_non_existing(self):
```
pcorpet, line 2285: Please add a test using a static newRoot: `'newRoot': {'document': 'new'}`.

tests/test__collection_api.py
```diff
@@ -2180,6 +2180,69 @@ def test__find_or_and(self):
...
+        self.db.a.insert_many([
```
pcorpet, line 2286: Please add a test, with the same data but with the config `'newRoot': {'dogs': '$pets.dogs'}`. The result should be:
```py
[{'dogs': 2}, {}]
```

pcorpet (changes requested):

Missing_Features.rst
ymoran00: Don't know how this change got in...

tests/test__collection_api.py
ymoran00: Do we need to write the exact error the mongo server writes? In the other locations in te code the errors were written differently. Doesn't it only matter that it will fail in the same places?

tests/test__collection_api.py
ymoran00: It seems that my code doesn't solve it, and it causes me some issues... I've added to the Missing_features document that the dot reference support is missing for replaceRoot.

Missing_Features.rst
```diff
@@ -21,8 +21,8 @@ If I miss to include a feature in the below list, Please feel free to add to the
   * `$indexStats <https://docs.mongodb.com/manual/reference/operator/aggregation/indexStats/>`_
   * `$listLocalSessions <https://docs.mongodb.com/manual/reference/operator/aggregation/listLocalSessions/>`_
   * `$listSessions <https://docs.mongodb.com/manual/reference/operator/aggregation/listSessions/>`_
-  * `$replaceRoot <https://docs.mongodb.com/manual/reference/operator/aggregation/replaceRoot/>`_
   * `$sortByCount <https://docs.mongodb.com/manual/reference/operator/aggregation/sortByCount/>`_
+  * `$replaceRoot <https://docs.mongodb.com/manual/reference/operator/aggregation/replaceRoot/>`_ - missing dot reference support.
```
pcorpet: Keep it sorted, or don't forget to remove if my code below works.

tests/test__collection_api.py
pcorpet: The trouble is not only in the dot:

If you add an extra data where `pets` field is missing then your code would fail. Note that you can fix (probably both by using):

```py
results = []
for doc in in_collection:
  try:
    results.append(_parse_expression(new_root, doc)
  except KeyError:
    results.appends({})
return results
```

What do you think?

tests/test__collection_api.py
pcorpet: I'm OK not to have the same text message. However it should be the same kind of exception (OperationFailure) and explain to the user what went wrong. Here I like the reminder that it must evaluate to an object.

pcorpet (changes requested):

tests/test__collection_api.py
ymoran00: That was my first hunch so I already tested it. The problem is that we get 2 types of KeyError here: one for the case we discuss here (dot reference), and one for the case we've discussed in test test__aggregate_replace_root_non_existing.
So If I catch the KeyError here - test__aggregate_replace_root_non_existing will fail...

pcorpet: You're right, I hadn't tried it and if `pets` field is missing in one doc then the query is crashing... I don't really like the current state because we're getting an OperationFailure in cases where we should get a NotImplementedError or the actual solution.

The changes would need to be larger. I think that adding a parameter to _parse_expression to ignore key errors would enable to make the difference. Let me try to do that and then you can rebase on my work.

pcorpet: I've done the rebasing myself. I'll let you have a look and merge if you're OK

pcorpet (approved):

PR #549, opened 2019-07-19
Handle array subfields in the  stage of the aggregate pipeline.

Mocramis:
While testing some $slice on aggregations, in noticed, arrays subfields inclusions were ignored.

Here is a fix :)

### DIFF
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -2403,6 +2403,25 @@ class MongoClientAggregateTest(_CollectionComparisonTest):
         pipeline = [{'$project': {'_id': 0}}]
         self.cmp.compare_ignore_order.aggregate(pipeline)
 
+    def test__aggregate_project_array_subfield(self):
+        self.cmp.do.insert_many([
+            {'a': [{'b': 1, 'c': 2, 'd': 3}], 'e': 4},
+            {'a': [{'c': 12, 'd': 13}], 'e': 14},
+            {'a': [{'b': 21, 'd': 23}], 'e': 24},
+            {'a': [{'b': 31, 'c': 32}], 'e': 34},
+            {'a': [{'b': 41}], 'e': 44},
+            {'a': [{'c': 51}], 'e': 54},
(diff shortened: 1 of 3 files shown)


tests/test__mongomock.py
```diff
@@ -2403,6 +2403,25 @@ def test__aggregate_project_id_0(self):
...
+            {'a': [{'b': 41}], 'e': 44},
+            {'a': [{'c': 51}], 'e': 54},
+            {'a': [{'d': 51}], 'e': 54},
+            {'a': []},
+            {'a': [1, 2, 3, 4]},
+            {'a': 'foobar'},
+            {'a': 5},
+        ])
+        pipeline = [
+            {'$project': {'a.b': 1, 'a.c': 1, '_id': 0}}
```
pcorpet: This should all fit on one line:
`pipeline = [{'$project': ...}]`

tests/test__mongomock.py
```diff
@@ -2403,6 +2403,25 @@ def test__aggregate_project_id_0(self):
         pipeline = [{'$project': {'_id': 0}}]
         self.cmp.compare_ignore_order.aggregate(pipeline)
 
+    def test__aggregate_project_array_subfield(self):
+        self.cmp.do.insert_many([
+            {'a': [{'b': 1, 'c': 2, 'd': 3}], 'e': 4},
```
pcorpet, line 2408: Please add one test with an array with a mix of dicts and int  `{'a': [{'b': 1}, 1, {'c': 2}, {'d': 3}, 'c']}`

pcorpet (changes requested):

tests/test__mongomock.py
Mocramis: indeed, i even forgot to test the multiple matching elements in an array case.

Mocramis:
@pcorpet  Is anything missing for merge request ? 
Thanks.

pcorpet (approved):

PR #550, opened 2019-07-22
Consider "name" argument to create_index for index name

rbu:
Current version does not allow giving a human readable index name.


schizef:
Thanks for the PR, I also need this!  However, it looks like `create_indexes` probably needs `name` added as well.

pcorpet (changes requested):
Yes indeed, please add name in `create_indexes` before I merge.

rbu:
> Yes indeed, please add name in `create_indexes` before I merge.

Thank you for the review. I've implemented this change as proposed.

pcorpet (approved):

PR #559, opened 2019-08-23
Bytes sort

Mocramis:
I noticed i couldn't  sort operations on bytes field.

I do not know if the tests are enough, but this is a quite simple patch.

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -8,7 +8,7 @@ import numbers
 import operator
 import re
 from sentinels import NOTHING
-from six import iteritems, iterkeys, string_types
+from six import iteritems, iterkeys, string_types, PY3
 try:
     from types import NoneType
 except ImportError:
@@ -287,6 +287,14 @@ def bson_compare(op, a, b, can_compare_types=True):
     if isinstance(a, NoneType):
         return op(0, 0)
 
+    # bson handles bytes as binary in python3+:
+    # https://api.mongodb.com/python/current/api/bson/index.html
+    if PY3 and isinstance(a, bytes):
+        # Performs the same operation as described by:
+        # https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#bindata
+        if len(a) != len(b):
+            return op(len(a), len(b))
+        # bytes is always treated as subtype 0 by the bson library
     return op(a, b)
 
 
@@ -309,6 +317,9 @@ def _get_compare_type(val):
         return 20
     if isinstance(val, (tuple, list)):
         return 25
+    if isinstance(val, bytes):
+        assert PY3
+        return 30
     if isinstance(val, ObjectId):
         return 35
     if isinstance(val, datetime):
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -3740,9 +3740,12 @@ class CollectionAPITest(TestCase):
             {'_id': 11, 'counts': {'circles': mongomock.ObjectId()}},
             # Document kept: datetimes are more than numbers.
             {'_id': 12, 'counts': {'circles': datetime.now()}},
+            # Document kept: BinData are more than numbers.
+            {'_id': 13, 'counts': {'circles': b'binary'}},
         ])
         results = collection.find(query)
-        self.assertEqual({1, 2, 5, 9, 11, 12}, {doc['_id'] for doc in results})
+        self.assertEqual(
+            {1, 2, 5, 9, 11, 12, 13}, {doc['_id'] for doc in results})
 
         query = {'counts': {'$gt': {'circles': re.compile('3')}}}
         self.assertFalse(list(collection.find(query)))
--- a/tests/diff.py
+++ b/tests/diff.py
@@ -11,7 +11,7 @@ class _NO_VALUE(object):
 NO_VALUE = _NO_VALUE()
 
 _SUPPORTED_TYPES = (float, bool, str, datetime.datetime, type(None)) + \
-    string_types + integer_types + (text_type,)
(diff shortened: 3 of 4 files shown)


mongomock/filtering.py
```diff
@@ -401,11 +405,24 @@ def operator_eq(doc_val, search_val):
     return operator.eq(doc_val, search_val)
 
 
+def cmp_operator(op, a, b):
```
pcorpet: Please treat this as a special case in the bson_compare function. Or did I miss something?

tests/test__collection_api.py
```diff
@@ -3740,9 +3740,12 @@ def test__filter_objects_comparison(self):
             {'_id': 11, 'counts': {'circles': mongomock.ObjectId()}},
             # Document kept: datetimes are more than numbers.
             {'_id': 12, 'counts': {'circles': datetime.now()}},
+            # Document kept: BinData are more than numbers.
+            {'_id': 13, 'counts': {'circles': b'binary'}},
         ])
         results = collection.find(query)
-        self.assertEqual({1, 2, 5, 9, 11, 12}, {doc['_id'] for doc in results})
+        self.assertEqual({1, 2, 5, 9, 11, 12, 13},
```
pcorpet: Please break line after `(`

mongomock/filtering.py
```diff
@@ -401,11 +405,24 @@ def operator_eq(doc_val, search_val):
...
+    # pymongo handles bytes as binary in python3+:
```
pcorpet: It's actually bson, not pymongo

pcorpet (changes requested):
Very cool! Thanks a lot for this. And thanks for all the pointer to the code or documentation.

mongomock/filtering.py
Mocramis: this only apply to sorting operators. appart from checkimg wether an operator is one of the sorting operator in bson_compare, i'm not sure how to do it, and the former method does not seem better to me than having the correct operators right offt the bat.

mongomock/filtering.py
Mocramis: my bad

mongomock/filtering.py
pcorpet: I'm not sure to understand what you're implying. `bson_compare` is only ever called with values from the `SORTING_OPERATOR_MAP` dict.

I think that comparing two bytes fields even when not sorting should use your method. `{'$lt': [b'abc', b'b']}`.

I suggest you keep the tests you have added and try moving the code to bson_compare: they should pass and not break any other tests.

Mocramis: my bad dunno why, i though it was called for other operators. done.

mongomock/filtering.py
```diff
@@ -1,4 +1,5 @@
 from datetime import datetime
+from functools import partial
```
pcorpet: Unused.

mongomock/filtering.py
```diff
@@ -287,6 +288,16 @@ def bson_compare(op, a, b, can_compare_types=True):
...
         return op(0, 0)
 
+    # bson handles bytes as binary in python3+:
+    # https://api.mongodb.com/python/current/api/bson/index.html
+    if PY3 and isinstance(a, bytes):
+        # Performs the same operation as described by:
+        # https://docs.mongodb.com/manual/reference/bson-type-comparison-order/#bindata
+        if len(a) != len(b):
+            return op(len(a), len(b))
+        # bytes is always treated as subtype 0 by the bson library
```
pcorpet, line 297: Drop those two lines, as the default is to run it in this function (see line 301).

Mocramis: nope if  i remove these, the ordering differs with the mongo one (with a slight change to the tests that I am going to push).

Mocramis: with current tests , i get (without these two lines):

```
======================================================================
FAIL: test__find_sort_list_empty_order (tests.test__mongomock.MongoClientCollectionTest)
----------------------------------------------------------------------
  File "/data/mongomock/tests/test__mongomock.py", line 829, in test__find_sort_list_empty_order
    self.cmp.compare.find(sort=[('b', 1)])
  File "/data/mongomock/tests/multicollection.py", line 70, in __call__
    _assert_no_diff(results, ignore_order=self.___ignore_order, sort_by=self.___sort_by)
  File "/data/mongomock/tests/multicollection.py", line 95, in _assert_no_diff
    assert not d, _format_diff_message(prev_name, name, d)
AssertionError: Unexpected Diff:
	fake.3.b != real.3.b (b'alongbin1' != b'bin1')
	fake.4.b != real.4.b (b'alongbin2' != b'bin2')
	fake.5.b != real.5.b (b'bin1' != b'alongbin1')
	fake.6.b != real.6.b (b'bin2' != b'alongbin2')
```

pcorpet: My point is that if you remove `'return op(a, b)`, Python will leave the `if PY3 and isinstance(a, bytes)` branch and go to the last line of the function which is also `return op(a, b)`, so I don't see how it goes wrong.

pcorpet: Basically:

```py
if a:
  if b:
     return 0
  return 1
return 1
```

is equivalent to 

```py
if a:
  if b:
     return 0
return 1
```

Mocramis: Ha ok. Dunno if it is due to an update, but the quoted code included the two lines before (and not the return) so i though you were talking about the 
```
if len(a) != len(b):
   return op(len(a), len(b))
```
lines

Sorry, should have guessed it. Will do !

pcorpet (approved):

PR #560, opened 2019-09-06
Support bson.regex on $not queries

vainu-jussi:
For example: collection.find_one({"field": {"$not": Regex("^searchstr")}})

Issue #561


pcorpet (approved):
Extra! Thanks a lot.

PR #581, opened 2019-12-17
Add support for subtracting milliseconds from a date

### LINKED ISSUE #580: $subtract doesn't support subtracting milliseconds from a date
This sort of aggregation pipeline groups documents by 5 minute intervals:

```python
[{
    '$group': {
        '_id': {
            'timestamp': {
                '$subtract' : ['$timestamp', {
                    '$mod': [{
                        '$subtract' : ['$timestamp', an_example_datetime]
                     }, 5 * 60 * 1000]
                }]
            }
        }
    }
]}
```

But it results in this exception:

```
TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'float'
```

Because the `$mod` resolves to a `float`, but `$timestamp` is a datetime.

It works however when run within MongoDB.

See the documentation here:

https://docs.mongodb.com/manual/reference/operator/aggregation/subtract/#subtract-milliseconds-from-a-date

PR incoming.

danielcrisp:
Fixes #580

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -240,7 +240,8 @@ class _Parser(object):
             assert len(values) == 2, 'subtract must have only 2 items'
             value_0 = self.parse(values[0])
             value_1 = self.parse(values[1])
-            if isinstance(value_0, datetime.datetime) and isinstance(value_1, (int, float)):
+            if isinstance(value_0, datetime.datetime) and \
+                    isinstance(value_1, (six.integer_types, float)):
                 value_1 = datetime.timedelta(milliseconds=value_1)
             res = value_0 - value_1
             if isinstance(res, datetime.timedelta):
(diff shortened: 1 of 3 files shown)


mongomock/aggregate.py
```diff
@@ -238,7 +238,11 @@ def _handle_arithmetic_operator(self, operator, values):
             return math.sqrt(self.parse(values))
         if operator == '$subtract':
             assert len(values) == 2, 'subtract must have only 2 items'
-            res = self.parse(values[0]) - self.parse(values[1])
+            value_0 = self.parse(values[0])
+            value_1 = self.parse(values[1])
+            if isinstance(value_0, datetime.datetime) and isinstance(value_1, (int, float)):
```
pcorpet: Let's not forget about `long` in Python 2. I suggest you use integer_types https://six.readthedocs.io/#six.integer_types

pcorpet (approved):
Very cool, thanks.

Please remove the `subtract` line from the `Missing_features` file as it seems pretty much implemented now.

mongomock/aggregate.py
danielcrisp: Awesome. Done.

Issue #585, opened 2020-01-16
collection.Collection.document_count() returns negative value


mgalka:
When the `skip` parameter passed to the `collection.Collection.document_count()` is greater than a number of documents retrieved from db the mocked method returns negative value, while the one from `pymongo` returns 0.

Tested on PyMongo 3.10.1 and Mongomock 3.18.0

Example:

**Running the code on a local MongoDB instance:**

```python
import pymongo


def do_something(client):
    db = client['testdb']
    col = db['testcol']
    doc = {'key1': 'value1', 'key2': 'value2'}
    col.insert_one(doc)
    return col.count_documents({}, skip=10)


if __name__  == '__main__':
    client = pymongo.MongoClient("mongodb://localhost:27017/")
    print(do_something(client))
```

This code prints 0

**Running a test using mongomock:**

```python
import mongomock


def do_something(client):
    db = client['testdb']
    col = db['testcol']
    doc = {'key1': 'value1', 'key2': 'value2'}
    col.insert_one(doc)
    return col.count_documents({}, skip=10)


def test_something():
    client = mongomock.MongoClient()
    assert do_something(client) == 0
```

The test fails:

```
   def test_something():
        client = mongomock.MongoClient()
>       assert do_something(client) == 0
E       AssertionError: assert -9 == 0
E        +  where -9 = do_something(mongomock.MongoClient('localhost', 27017))

test_mongo_skip.py:12: AssertionError
```

pcorpet:
Ha, nice catch. Would you care to send a PR please?

mgalka:
PR #586 created to fix this issue

PR #586, opened 2020-01-16
collection.py: Fix return value of Collection.documents_count()

mgalka:
When skip was greater than count for
collection.Collection.document_count() the method kept returning
negative values It returns 0 now, which is the expected behaviour.

Fixes Issue #585

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1357,10 +1357,9 @@ class Collection(object):
         if unknown_kwargs:
             raise OperationFailure("unrecognized field '%s'" % unknown_kwargs.pop())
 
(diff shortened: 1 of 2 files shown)


pcorpet (changes requested):
Thanks for your change, please add a test in test__mongomock so we do not have regressions in the future.

mgalka:
> Thanks for your change, please add a test in test__mongomock so we do not have regressions in the future.

I added the tests.

pcorpet (approved):

pcorpet:
Thanks for the quick turnaround on that issue.

PR #588, opened 2020-01-21
Support for $slice operator in projection argument

kdeyev:
Hi,

I've added support for lists slicing in the projection argument. Like that:
find({'name': 'Array'}, projection={'name': 1, 'values': {'$slice': [-3, 10]}}
[Mongodb documentation](https://docs.mongodb.com/manual/reference/operator/projection/slice/)

Please, review and approve :-)

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1037,7 +1037,7 @@ class Collection(object):
                     slice_ = None        
                     if isinstance(op_value, list):
                         if len(op_value) != 2:
-                            raise ValueError('Unsupported slice format {} for slicing operation: {}'.format(op_value, op))
+                            raise OperationFailure('Unsupported slice format {} for slicing operation: {}'.format(op_value, op))
                         skip = op_value[0]
                         limit = op_value[1]
                         if skip < 0:
@@ -1057,9 +1057,9 @@ class Collection(object):
                     if slice_:
                         doc_copy[field] = doc_copy[field][slice_]
                     else:
-                        raise ValueError('Unsupported slice value {} for slicing operation: {}'.format(op_value, op))
+                        raise OperationFailure('Unsupported slice value {} for slicing operation: {}'.format(op_value, op))
                 else:
-                    raise ValueError('Unsupported type {} for slicing operation: {}'.format(type(doc_copy[field]), op))
+                    raise OperationFailure('Unsupported type {} for slicing operation: {}'.format(type(doc_copy[field]), op))
 
             if '$elemMatch' in op:
                 if isinstance(doc_copy[field], list):
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1038,7 +1038,8 @@ class Collection(object):
                     if isinstance(op_value, list):
                         if len(op_value) != 2:
                             raise OperationFailure(
-                                'Unsupported slice format {} for slicing operation: {}'.format(op_value, op))
+                                'Unsupported slice format {} for slicing operation: {}'.format(
+                                    op_value, op))
                         skip = op_value[0]
                         limit = op_value[1]
                         if skip < 0:
@@ -1059,10 +1060,12 @@ class Collection(object):
                         doc_copy[field] = doc_copy[field][slice_]
                     else:
                         raise OperationFailure(
-                            'Unsupported slice value {} for slicing operation: {}'.format(op_value, op))
+                            'Unsupported slice value {} for slicing operation: {}'.format(
+                                op_value, op))
                 else:
                     raise OperationFailure(
-                        'Unsupported type {} for slicing operation: {}'.format(type(doc_copy[field]), op))
+                        'Unsupported type {} for slicing operation: {}'.format(
+                            type(doc_copy[field]), op))
 
             if '$elemMatch' in op:
                 if isinstance(doc_copy[field], list):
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -1031,6 +1031,42 @@ def _apply_projection_operators(self, ops, doc, doc_copy):
...
+                        slice_ = slice(start, end)
+
+                    if slice_:
+                        doc_copy[field] = doc_copy[field][slice_]
+                    else:
+                        raise OperationFailure(
+                            'Unsupported slice value {} for slicing operation: {}'.format(
+                                op_value, op))
+                else:
+                    raise OperationFailure(
```
pcorpet, line 1064: I suggest you reverse your if so that you can raise and avoid an `else` branch:
```py
if not is instance(doc_copy[field], list):
  raise ...
op_value = op['$slice']
```

mongomock/collection.py
```diff
@@ -1031,6 +1031,42 @@ def _apply_projection_operators(self, ops, doc, doc_copy):
...
+            if '$slice' in op:
+                if isinstance(doc_copy[field], list):
+                    op_value = op['$slice']
+                    slice_ = None
+                    if isinstance(op_value, list):
+                        if len(op_value) != 2:
+                            raise OperationFailure(
+                                'Unsupported slice format {} for slicing operation: {}'.format(
+                                    op_value, op))
+                        skip = op_value[0]
```
pcorpet:
```py
skip, limit = op_value
```

tests/test__mongomock.py
```diff
@@ -1001,6 +1001,64 @@ def test__default_fields_to_id_if_empty(self):
...
+        self.cmp.compare.find({'name': 'Array'}, projection={
+            'name': 1, 'values': {'$slice': [-3, 10]}})
+
+    def test__projection_slice_list_select_subfield(self):
+        self.cmp.do.insert({'name': 'Array', 'values': [
+            {'num': 0, 'val': 1}, {'num': 1, 'val': 2}]})
+        self.cmp.compare.find({'name': 'Array'}, projection={
+            'values.num': 1, 'values': {'$slice': 1}})
+
+    def test__projection_slice_list_wrong_num_slice(self):
```
pcorpet, line 1040: Please move those tests to test__collection_api as the compare object actually does two operations (one on a real Mongo DB, the other on a mongomock) but the exception would raise only for the first one, thus it's not testing what you wish. I'll work on a way to make that easier.

pcorpet: OK done. Rebase then use `self.cmp.compare_exceptions.find({'name': ...`

pcorpet (changes requested):

mongomock/collection.py
kdeyev: Done

kdeyev: Done

tests/test__mongomock.py
kdeyev: I've started to use `compare_exceptions` method. Thank you!
I looked at test__collection_api file and I see that no one of tests is using `MultiCollection` for comparison of results. Actually it means that moving of new tests will require to bring some additional code to the `test__collection_api` (such as `_CollectionComparisonTest` for example). 
Please let me know if you want me anyway moving the new tests into `test__collection_api`

mongomock/collection.py
kdeyev: Done

kdeyev (commented):
Thank you for reviewing my PR. 
I've followed the CR comments. Please review the new commits.

tests/test__mongomock.py
pcorpet: Nope, it's OK now that we have `compare_exceptions`.

tests/test__mongomock.py
```diff
@@ -1003,6 +1003,55 @@ def test__default_fields_to_id_if_empty(self):
...
+        self.cmp.do.insert({'name': 'Array', 'values': [0, 1, 2, 3, 4, 5, 6, 7]})
+        self.cmp.compare_exceptions.find({'name': 'Array'}, projection={
+            'name': 1, 'values': {'$slice': [-3, 10, 1]}})
+
+    def test__projection_slice_list_wrong_slice_type(self):
+        self.cmp.do.insert({'name': 'Array', 'values': [0, 1, 2, 3, 4, 5, 6, 7]})
+        self.cmp.compare_exceptions.find({'name': 'Array'}, projection={
+            'name': 1, 'values': {'$slice': [1.0]}})
+
+    def test__projection_slice_list_wrong_value_type(self):
```
pcorpet, line 1055: Please add one last test with a `slice` using neither a list nor an int: `{'name': 1, 'values': {'$slice': '3'}}`

pcorpet (changes requested):
Thanks for all the change. It's almost good to go.

tests/test__mongomock.py
kdeyev: Done

kdeyev (commented):
One more test was added. Please, approve.

pcorpet (approved):

pcorpet:
Thanks for the very clean result @kdeyev

kdeyev:
Thank you! If you don't mind I'll continue to add MongoDB features support.

PR #578, opened 2019-12-03
added code for $dateToString operation

SushmithaMiriyam:

PR #590, opened 2020-01-25
Add $dateToString opeartor

creek0810:
add basic datetostring operator

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -359,9 +359,31 @@ class _Parser(object):
         if operator == '$millisecond':
             return int(out_value.microsecond / 1000)
         if operator == '$dateToString':
+            if not isinstance(values, dict):
+                raise OperationFailure(
+                    '$dateToString operator must correspond a dict'
+                    'that has "format" and "date" field.'
+                )
+            if 'format' not in values.keys() or 'date' not in values.keys():
+                raise OperationFailure(
+                    '$dateToString operator must correspond a dict'
+                    'that has "format" and "date" field.'
+                )
             if '%L' in out_value['format']:
                 raise NotImplementedError(
                     'Although %L is a valid date format for the '
+                    '$dateToString operator, it is currently not implemented '
+                    ' in Mongomock.'
+                )
+            if 'onNull' in values.keys():
+                raise NotImplementedError(
+                    'Although onNull is a valid field for the '
+                    '$dateToString operator, it is currently not implemented '
+                    ' in Mongomock.'
+                )
+            if 'timezone' in values.keys():
+                raise NotImplementedError(
+                    'Although timezone is a valid field for the '
                     '$dateToString operator, it is currently not implemented '
                     ' in Mongomock.'
                 )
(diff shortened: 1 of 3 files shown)


mongomock/aggregate.py
```diff
@@ -358,6 +358,15 @@ def _handle_date_operator(self, operator, values):
...
         if operator == '$millisecond':
             return int(out_value.microsecond / 1000)
+        if operator == '$dateToString':
+            if '%L' in out_value['format']:
+                raise NotImplementedError(
+                    'Although %L is a valid date format for the '
+                    '$dateToString operator, it is currently not implemented '
+                    ' in Mongomock.'
+                )
+            return out_value['date'].strftime(out_value['format'])
```
pcorpet, line 390: What if `date` or `format` parameter is missing?

mongomock/aggregate.py
```diff
@@ -358,6 +358,15 @@ def _handle_date_operator(self, operator, values):
             return out_value.second
         if operator == '$millisecond':
             return int(out_value.microsecond / 1000)
+        if operator == '$dateToString':
+            if '%L' in out_value['format']:
```
pcorpet, line 372: Please also add `NotImplementedError` when using `onNull` or `timezone` parameters.

pcorpet (changes requested):
Please check #578, there are some comments that are still valid here.

mongomock/aggregate.py
creek0810: done

mongomock/aggregate.py
creek0810: add code to raise OperationFailure

creek0810:
have added some code to handle OperationFailure and NotImplementedError condition

mongomock/aggregate.py
```diff
@@ -358,6 +358,37 @@ def _handle_date_operator(self, operator, values):
             return out_value.second
         if operator == '$millisecond':
             return int(out_value.microsecond / 1000)
+        if operator == '$dateToString':
+            if not isinstance(values, dict):
+                raise OperationFailure(
+                    '$dateToString operator must correspond a dict'
+                    'that has "format" and "date" field.'
+                )
+            if 'format' not in values.keys() or 'date' not in values.keys():
```
pcorpet: Let's move that above:
```py
if not isinstance(values, dict) or not {'format', 'date'} <= set(values.keys()):
```

pcorpet: Actually `set(values)` instead of `set(values.keys())`

mongomock/aggregate.py
```diff
@@ -358,6 +358,37 @@ def _handle_date_operator(self, operator, values):
...
+                    '$dateToString operator must correspond a dict'
+                    'that has "format" and "date" field.'
+                )
+            if '%L' in out_value['format']:
+                raise NotImplementedError(
+                    'Although %L is a valid date format for the '
+                    '$dateToString operator, it is currently not implemented '
+                    ' in Mongomock.'
+                )
+            if 'onNull' in values.keys():
```
pcorpet: No need to check `.keys()`, just `if 'onNull' in values:`

pcorpet (changes requested):
Only cosmetic changes requested.

mongomock/aggregate.py
creek0810: Done

mongomock/aggregate.py
creek0810: Done

PR #596, opened 2020-02-19
Implements $setEquals operator

lucasrcezimbra:


tests/test__collection_api.py
```diff
@@ -3454,11 +3454,6 @@ def test__aggregate_not_implemented(self):
                 {'$project': {'a': {'$concatArrays': [[0, 1], [2, 3]]}}},
             ])
 
-        with self.assertRaises(NotImplementedError):
-            self.db.collection.aggregate([
-                {'$project': {'a': {'$setEquals': [[2], [1, 2, 3]]}}},
```
pcorpet, line 3537: Please keep this one but with `setIntersection` to keep code coverage high.

tests/test__collection_api.py
```diff
@@ -4682,6 +4677,32 @@ def test__set_union(self):
         }]
         self.assertEqual(expect, list(actual))
 
+    def test__set_equals(self):
+        collection = self.db.collection
+        collection.insert_many([
+            {'array': ['one', 'three']},
+        ])
+        actual = collection.aggregate([{'$project': {
```
pcorpet, line 4800: Please also add an equivalent case in test__mongomock to make sure the behavior is the same than the one of pymongo.

pcorpet (changes requested):
Thanks a lot for implementing the feature.

Let's make sure that all CI checks are successful.

tests/test__collection_api.py
lucasrcezimbra: Done!

tests/test__collection_api.py
lucasrcezimbra: Done!

pcorpet (approved):

PR #601, opened 2020-03-05
Drop the test database rather than the test collection

oorijsonar:
Since one of the tests rename a collection (test__rename_collection) and nothing cleans it up, this test fails on the second run (`pymongo.errors.OperationFailure: target namespace exists`).


oorijsonar:
Any idea how I managed to drop the coverage?

PR #605, opened 2020-03-11
Project missing fields in a nested dict fix

superqwer:
Current behavior of mongomock is "Ignore the whole dictionary if we couldn't find a single key"

That's not what the real mongo does

The real mongo just ignores this missing key
```
> db.asdf.aggregate([{'$project': {'a': 1, 'b': {'c': '$b'}}}])
{ "_id" : ObjectId("5e6875c3c3b830d2aaf259e5"), "a" : 1, "b" : {  } }
```

### DIFF
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -3068,6 +3068,17 @@ class CollectionAPITest(TestCase):
         ])
         self.assertEqual([{}], list(actual))
 
+    def test__aggregate_project_missing_nested_fields(self):
+        self.db.collection.insert_one({'_id': 1, 'a': 2, 'b': {'c': 1}})
+        actual = self.db.collection.aggregate([
+            {'$match': {'_id': 1}},
+            {'$project': collections.OrderedDict([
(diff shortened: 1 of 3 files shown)


tests/test__collection_api.py
```diff
@@ -3068,6 +3068,17 @@ def test__aggregate_project_missing_fields(self):
         ])
         self.assertEqual([{}], list(actual))
 
+    def test__aggregate_project_missing_nested_fields(self):
```
pcorpet, line 3071: Please add an equivalent test in `test__mongomock` to ensure the behavior is kept the same between `pymongo` and `mongomock`

pcorpet (changes requested):
Thanks for finding the bug and fixing it.

superqwer:
Added similar testcase to test_mongomock

tests/test__collection_api.py
superqwer: done

pcorpet (approved):

PR #606, opened 2020-03-16
$pushall $addToSet dollar operand support

superzarzar:
These operations do not work without the suggested changes:
1. {"$pullAll": {"takes.$.tags": [1, 2, 3]}}
2. {"$addToSet": {"takes.$.tags": 3}}

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -655,6 +655,15 @@ class Collection(object):
                             continue
                         # push to array in a nested attribute
                         else:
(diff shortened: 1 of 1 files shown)


pcorpet (changes requested):
Thanks for providing a fix.

Please add a test demonstrating the usecase you're trying to fix. Maybe you did in an intermediate commit but I don't see it in the final change (all commits at once).

pcorpet:
I'm taking over in #786, thanks @superzarzar

PR #612, opened 2020-04-01
Fix the not exist case in $toInt and $toDecimal

lxy1992:
It's the fix for the bug I mentioned in the last PR. 

First, I thought it was a serious bug in `_Parse` function, so I haven't fix it in the last PR. Now, I believe it only affect the operators in the `$addField` operator. So, I fix them in this PR with UTs.

The case is, when we try to add a field from a field which is not exist, Mongo will return the new field with the value of None.

Now, the `$toInt` and `$toDecimal` operators will return the original query now. So I return None when the `KeyError` raised in these operators.

If you're convenient, please review and merge it. 

Cheers!

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -481,7 +481,10 @@ class _Parser(object):
             return str(parsed)
 
         if operator == '$toInt':
-            parsed = self.parse(values)
+            try:
+                parsed = self.parse(values)
+            except KeyError:
+                return None
             if decimal_support:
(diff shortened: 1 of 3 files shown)


lxy1992:
I am working on another PR about $sum. I may finish it at this weekend. Looking forward your code review, thanks!

pcorpet (approved):
Perfect, thanks.

Feel free to send a PR to add yourself as a contributor to this project in the main page.

lxy1992:
OK, I will add this info in the next PR. Thanks!

PR #617, opened 2020-04-05
Support sum decimal type of values in $sum operator

lxy1992:
This PR add the support for the decimal value in `$sum` operator with UTs

Behavior:
* any value in the list of `$sum` operator is `decimal128.Decimal`, the result of sum will be `decimal128.Decimal128`.

The mongo support all types in `numbers.Number`, the `decimal` is one of them. However, the real type of decimal mongo store in the disk is `decimal128.Decimal128`, which is not supported by the `sum` function in Python. So, I convert them to `decimal.Decimal` before sum, revert the type of result when return.

Looking forwarding to code review.

Cheers!

### DIFF
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -4194,7 +4194,7 @@ class CollectionAPITest(TestCase):
             list(actual)
         )
 
-    def test_aggregate_group_sum(self):
+    def test__aggregate_group_sum(self):
         collection = self.db.collection
         collection.insert_many([
             {'group': 'one'},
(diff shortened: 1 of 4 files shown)


tests/test__collection_api.py
```diff
@@ -4224,6 +4224,32 @@ def test_aggregate_group_sum(self):
         }]
         self.assertEqual(expect, list(actual))
 
+    @skipIf(not _HAVE_PYMONGO, 'pymongo not installed')
+    def test__aggregate_group_sum_for_decimal(self):
+        collection = self.db.collection
+        collection.drop()
+        collection.insert_one(
+            {
```
pcorpet: Please move this opening curly bracket together with the previous line. (Note that in general I try to reduce the indentation if there's only one value inside).

pcorpet (changes requested):
Thanks again.

One lint I didn't tell you about last time but I prefer to have in here if you don't mind.

tests/test__collection_api.py
lxy1992: It's reasonable.  I have reformed these codes. Thanks.

pcorpet (approved):

lxy1992:
I'm terribly sorry, I just find I use the wrong git author info for this commit. May I revert this PR and reopen one? @pcorpet

pcorpet:
I've reverted, please send me another one.

lxy1992:
> I've reverted, please send me another one.

@pcorpet 
Thank you so much, I have open a new PR in [https://github.com/mongomock/mongomock/pull/618](url)

PR #619, opened 2020-04-07
improve aggregation expression parsing

reyannh:
Make expressions starting with `$$ROOT` or `$$CURRENT` work on documents with array of embedded documents fields.
Consider the following document dict:
```{'name': 'foo', 'errors': [{'error_type': 1, 'description': 'problem 1'}, {'error_type': 2, 'description': 'problem 2'}], '_id': ObjectId('5e8c9ac81e957db62b16b888')}```
And the expression - `$$ROOT.errors.error_type`
Current code will throw KeyError exception - `{KeyError}2`

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -303,7 +303,7 @@ class _Parser(object):
                 return helpers.get_value_by_dot(dict({
                     'ROOT': self._doc_dict,
                     'CURRENT': self._doc_dict,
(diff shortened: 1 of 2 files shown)


pcorpet:
Thanks for suggesting this improvement. Please add some tests in test__mongomock, both to check that the behavior is the same as with pymongo + MongoDb and that your feature doesn't get broken by a later refactoring.

PR #623, opened 2020-04-22
Raise a WriteError when trying to update with an empty $set.

pcorpet:


tests/test__mongomock.py
```diff
@@ -1188,6 +1188,9 @@ def test__update_one(self):
                                upsert=True)
         self.cmp.compare.find()
 
+        self.cmp.compare_exceptions.update_one({}, {'$set': {}})
+        self.cmp.compare_exceptions.update_one({'a': 'does-not-exist'}, {'$set': {}})
```
zozoens31, line 1192: please, also test update_many

zozoens31, line 1192: please, also test on upsert (although upsert shouldn't be used with `$set`, maybe the output error is not the right one)

zozoens31 (changes requested):
LGTM

tests/test__mongomock.py
pcorpet: Done.

pcorpet: Done.

PR #624, opened 2020-05-06
Implement $arrayToObject operator

hugo19941994:
Hi!

I've implemented the arrayToObject operator for the aggregation pipeline.

I've added tests for all the edge cases I could think of:

* Mixed array and document inside the same array
* Extra keys in documents
* Missing keys in document
* Empty document
* More than two elements in an array
* Just one element in an array
* Empty array

Duplicated keys are also accounted for and tested (latest found value is used, overwriting previous values), but [there is a caveat in the mongo documentation for earlier versions](https://docs.mongodb.com/manual/reference/operator/aggregation/arrayToObject/#behavior). I'm not sure if this should be accounted for or mocking the behaviour for the latest versions is enough.

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -117,6 +117,7 @@ type_convertion_operators = [
     '$toString',
     '$toInt',
     '$toDecimal',
+    '$arrayToObject'
 ]
 
 
@@ -542,6 +543,32 @@ class _Parser(object):
             else:
                 raise TypeError("'%s' type is not supported" % type(parsed))
             return decimal_value
+
+        # Document: https://docs.mongodb.com/manual/reference/operator/aggregation/arrayToObject/
+        if operator == '$arrayToObject':
+            try:
+                parsed = self.parse(values)
+            except KeyError:
+                return None
+
+            if parsed is None:
+                return None
+
+            if not isinstance(parsed, (list, tuple)):
+                raise OperationFailure(
+                    '$arrayToObject requires an array input, found: {}'.format(type(parsed))
+                )
+
+            if all(isinstance(x, dict) and set(x.keys()) == {'k', 'v'} for x in parsed):
+                return {d['k']: d['v'] for d in parsed}
+
+            if all(isinstance(x, (list, tuple)) and len(x) == 2 for x in parsed):
+                return dict(parsed)
+
+            raise OperationFailure(
+                'arrays used with $arrayToObject must contain documents '
+                'with k and v fields or two-element arrays'
+            )
 
     def _handle_conditional_operator(self, operator, values):
         if operator == '$ifNull':
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -5153,3 +5153,103 @@ class CollectionAPITest(TestCase):
             self.db.collection.aggregate(
                 [{'$project': {'a': {'$dateToString': '10'}}}]
             )
+
+    def test__aggregate_array_to_object(self):
+        collection = self.db.collection
+        collection.insert_many([{
+            'items': [['a', 1], ['b', 2], ['c', 3], ['a', 4]]
+        }, {
+            'items': (['a', 1], ['b', 2], ['c', 3], ['a', 4])
+        }, {
+            'items': [('a', 1), ('b', 2), ('c', 3), ('a', 4)]
+        }, {
+            'items': (('a', 1), ('b', 2), ('c', 3), ('a', 4))
+        }, {
+            'items': [['a', 1], ('b', 2), ['c', 3], ('a', 4)]
+        }, {
+            'items': (['a', 1], ('b', 2), ['c', 3], ('a', 4))
+        }, {
+            'items': [{'k': 'a', 'v': 1}, {'k': 'b', 'v': 2},
+                      {'k': 'c', 'v': 3}, {'k': 'a', 'v': 4}],
+        }, {
+            'items': [],
+        }, {
+            'items': (),
+        }, {
+            'items': None,
+        }])
+
+        actual = collection.aggregate([
+            {'$project': {
+                'items': {'$arrayToObject': '$items'},
+                'not_exists': {'$arrayToObject': '$nothing'},
+                '_id': 0
+            }}
+        ])
+
+        expect = [{
+            'items': {'a': 4, 'b': 2, 'c': 3},
+            'not_exists': None
+        }, {
+            'items': {'a': 4, 'b': 2, 'c': 3},
+            'not_exists': None
+        }, {
+            'items': {'a': 4, 'b': 2, 'c': 3},
+            'not_exists': None
+        }, {
+            'items': {'a': 4, 'b': 2, 'c': 3},
+            'not_exists': None
+        }, {
+            'items': {'a': 4, 'b': 2, 'c': 3},
+            'not_exists': None
+        }, {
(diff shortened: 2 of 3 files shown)


hugo19941994:
I'll add some tests for those missing lines for coverage

hugo19941994:
It looks like `$arrayToObject` with a `null` value is valid too, and returns another `null`. I've added this edge case too.

tests/test__mongomock.py
```diff
@@ -3065,6 +3065,36 @@ def test_aggregate_date_to_string(self):
         ]
         self.cmp.compare.aggregate(pipeline)
 
+    def test_aggregate_array_to_object(self):
+        self.cmp.do.drop()
+        self.cmp.do.insert_one({
```
pcorpet: Please use insert_many to add all those docs at once.

tests/test__collection_api.py
```diff
@@ -5153,3 +5153,93 @@ def test__aggregate_date_to_string(self):
...
+        }, {
+            'items': {},
+            'not_exists': None
+        }, {
+            'items': None,
+            'not_exists': None
+        }]
+        self.assertEqual(expect, list(actual))
+
+        with self.assertRaises(TypeError):
```
pcorpet: Here you could use a table test: create a list of `items` that should trigger an error and then iterate over them and do an aggregation and an asserRaises for each of them.

tests/test__collection_api.py
```diff
@@ -5153,3 +5153,93 @@ def test__aggregate_date_to_string(self):
...
+            collection.aggregate([
+                {'$addFields': {'items': ''}},
+                {'$project': {'items': {'$arrayToObject': '$items'}, '_id': 0}}
+            ])
+
+        with self.assertRaises(TypeError):
+            collection.aggregate([
+                {'$addFields': {'items': 100}},
```
pcorpet: Add all those errors in test__mongomock using `compare_exceptions`

mongomock/aggregate.py
```diff
@@ -543,6 +544,41 @@ def _handle_type_convertion_operator(self, operator, values):
...
+        if operator == '$arrayToObject':
+            try:
+                parsed = self.parse(values)
+            except KeyError:
+                return None
+
+            if parsed is None:
+                return None
+
+            if not isinstance(parsed, list):
```
pcorpet: In general were `list` is supported, `tuple` is supported as well. Can you check on `pymongo` and update it here please?

mongomock/aggregate.py
```diff
@@ -543,6 +544,41 @@ def _handle_type_convertion_operator(self, operator, values):
...
+                raise TypeError('$arrayToObject only supports arrays')
+
+            def check_doc(elements):
+                return all([isinstance(x, dict) and 'k' in x and 'v' in x and len(x) == 2
```
pcorpet: I don't think you need the `[]` here; Also you might be able to check with set equality:
```py
return all(isinstance(x, dict) and x.keys() == {'k', 'v'} for x in elements)
```

mongomock/aggregate.py
```diff
@@ -543,6 +544,41 @@ def _handle_type_convertion_operator(self, operator, values):
...
+                            for x in elements])
+
+            def check_arr(elements):
+                return all([isinstance(x, list) and len(x) == 2 for x in elements])
```
pcorpet: here as well, please drop the `[]` and allow for `tuple`

mongomock/aggregate.py
```diff
@@ -543,6 +544,41 @@ def _handle_type_convertion_operator(self, operator, values):
...
+
+            if not any([check_doc(parsed), check_arr(parsed)]):
+                raise TypeError(
+                    'arrays used with $arrayToObject must contain documents with k and v fields '
+                    'or two-element arrays')
+
+            arr_to_obj = {}
+            for elements in parsed:
+                if isinstance(elements, list):
```
pcorpet: This one can just be built with `arr_to_obj = dict(parsed)`

pcorpet (changes requested):
Great works, thanks for adding.

hugo19941994:
Thanks for the thorough review and your suggestions. I've implemented your requests and squashed my commits together. In addition to your requests I've made the following changes:

- Implementation was further refactored
  - Used dictionary comprehension to convert `{k, v}` values to the expected objects
  - Dropped the two `defs`
  - Dropped the `any` and replaced by plain `if/elses`
- I've replaced the `TypeErrors` with PyMongo's `OperationFailure` (so that compare_exceptions worked)
- Added a bunch of tests to check that tuples work just as well as lists

While checking PyMongo I've noticed that following exception messages can be thrown:

```
$arrayToObject requires an array of size 2 arrays,found array of size: 3
$arrayToObject requires an object keys of 'k' and 'v'. Found incorrect number of keys:3
$arrayToObject requires an object with keys 'k' and 'v'. Missing either or both keys from: {v: 1, t: "t"}
$arrayToObject requires a consistent input format. Elements mustall be arrays or all be objects. Array was detected, now found: object
```

[All of the messages can be seen in the mongo repo](https://github.com/mongodb/mongo/blob/d169769bba283cd0a1906c81580ff8e7cf7bcd93/src/mongo/db/pipeline/expression.cpp#L613). My implementation doesn't check the reason for an OperationFailure in most cases and uses a [generic error message](https://github.com/mongomock/mongomock/pull/624/files#diff-a212319bf3834fb5cfa6ea8acbafba3fR567). If you want I can change the code so that the same messages are used, but I wanted to check with you before adding more complexity to the code.

EDIT: It looks like I broke something with Python 2.7. I'll take a look.

mongomock/aggregate.py
```diff
@@ -543,6 +544,29 @@ def _handle_type_convertion_operator(self, operator, values):
...
+            try:
+                parsed = self.parse(values)
+            except KeyError:
+                return None
+
+            if parsed is None:
+                return None
+
+            if not isinstance(parsed, (list, tuple)):
+                raise OperationFailure('$arrayToObject requires an array input, '
```
pcorpet: Please break after `OperationFailure(`: that helps with cleaner indentation that does not depend on the length of the class name.

mongomock/aggregate.py
```diff
@@ -543,6 +544,29 @@ def _handle_type_convertion_operator(self, operator, values):
...
+                                       'found: {}'.format(type(parsed)))
+
+            if all(isinstance(x, dict) and x.keys() == {'k', 'v'} for x in parsed):
+                return {d['k']: d['v'] for d in parsed}
+
+            if all(isinstance(x, (list, tuple)) and len(x) == 2 for x in parsed):
+                return dict(parsed)
+
+            raise OperationFailure('arrays used with $arrayToObject must contain documents '
```
pcorpet: Same here, break after `OperationFailure(`

pcorpet (changes requested):
Thanks for the neat changes..
* Totally OK not to have the same error message, as long as it's the same type. So let's keep it simpler.
* I prefer that we don't account for old versions of pymongo so your solution is best.

 Please try to fix the python 2.7 version and my 2 little comments on indentation then I'll merge.

hugo19941994:
Python 2.7 should be fixed now. `.keys()` returns a list in Python 2.7 which can't be directly compared against a set (unlike in python 3).

pcorpet (approved):

PR #633, opened 2020-05-27
added context mananger methods to mockmongo client

TravisHeppner:
The mockmongo client lacks the __enter__ and __exit__ methods that enables the use of 'with' syntax.
Simple pull request to add those methods.


pcorpet:
Perfect, thanks.

Issue #584, opened 2020-01-15
Aggregation breaks if $group receives DBRef as _id


AlexMurmour:
When we execute the aggregate command with DBRef as '_id', that raises a NotImplementedError exception.
```
pipeline = [
            {
                '$group': {
                    '_id': '$visit',
                    ...
                }
            },
        ]
result = db.collection.aggregate(pipeline)
```
Error message:
```
NotImplementedError: Mongomock does not know how to sort 'DBRef('visits', ObjectId('...'))' of type '<class 'bson.dbref.DBRef'>
```
However, it works with pymongo.

pcorpet:
Fixed by #636

PR #636, opened 2020-06-15
Add support for group by DBRefs

mstrotta:
Fixes issue #584

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -16,7 +16,7 @@ except ImportError:
     NoneType = type(None)
 
 try:
-    from bson import Regex
+    from bson import Regex, DBRef
     _RE_TYPES = (RE_TYPE, Regex)
 except ImportError:
     _RE_TYPES = (RE_TYPE,)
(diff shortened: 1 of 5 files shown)


mongomock/filtering.py
```diff
@@ -16,7 +16,7 @@
     NoneType = type(None)
 
 try:
-    from bson import Regex
+    from bson import Regex, DBRef
```
pcorpet, line 19: `mongomock` should work even if `pymongo` is not installed. Please fix what should happen in that case.

mongomock/filtering.py
```diff
@@ -350,6 +355,8 @@ def _get_compare_type(val):
         return 45
     if isinstance(val, _RE_TYPES):
         return 50
+    if isinstance(val, DBRef):
+        return 55
```
pcorpet: Thanks for adding it here as well, please make sure that the sorting of type is working as in `pymongo` by adding a line in `test__sort_mixed_types` in `test__mongomock.py`

tests/test__collection_api.py
```diff
@@ -4273,6 +4273,26 @@ def test__aggregate_group_dict_key(self):
             list(actual)
         )
 
+    def test__aggregate_group_dbref_key(self):
```
pcorpet, line 4329: Thanks for adding a unit test. Please also add an equivalent test in `test__mongomock.py` to ensure consistency with `pymongo` behavior.

pcorpet (changes requested):
Thanks a lot for providing a PR to fix that bug. Please fix Travis and take a look at my comment.

kosqx:
Hi! Your CI build failed because of the https://github.com/mongomock/mongomock/issues/648 error.

I will push the fix soon

kosqx:
The fix is ready, tests are passing: https://github.com/mongomock/mongomock/pull/649

PR #640, opened 2020-07-06
DOC: Clarify how to use mongomock.patch with pymongo

### LINKED ISSUE #632: [Question] Clarify the limitation of mongomock.patch
Hi I was trying to use `@mongomock.patch` but couldn't get it to work. The situation is that I am using the [mongodbio](https://github.com/apache/beam/blob/master/sdks/python/apache_beam/io/mongodbio.py) connector for Beam. It is this function that internally creates a mongodb connection. I can specify the uri it should use and patch that uri, but internally it cannot connect to that address, e.g. `pymongo.errors.ServerSelectionTimeoutError: fakehost.com:27017: timed out`. If I use `localhost`, then data will actually be written to the local database.

This made me think `@mongomock.patch` isn't working here, and I wonder if the reason is because the code creating the client (e.g. beam code calling pymongo.MongoClient) is not within the scope of my test and module, and there is no way to inject it.

MarcoGorelli:
closes #632

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -76,7 +76,7 @@ This code checks *increase_votes* with respect to its functionality, not syntax 
 therefore is much more robust as a test.
 
 If the code to be tested is creating the connection itself with pymongo, you can use
-mongomock.patch:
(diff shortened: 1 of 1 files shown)


pcorpet (approved):
Thanks.

PR #643, opened 2020-07-13
doc_copy[key] = copy.deepcopy(val)

Bezbran:
Look in this case:
first_call = mongo_c.get_collection("collectionA").find_one({"search_field": "search_value"}, projection={"som_field":False})
second_call = mongo_c.get_collection("collectionA").find_one({"search_field": "search_value"}, projection={"som_field":False})

and there is an item with {"search_field": "search_value","som_field":"abc","some_field2":{"inner_field1":"original value"}}.

if we will do something like:
first_call ["some_field2"]["inner_field1"] = "new_value"
we will change the value of the second_call  too:
print(second_call ["some_field2"]["inner_field1"] )
>new_value
**which is wrong!!!**

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -17,6 +17,7 @@ from six import moves, raise_from
 from mongomock import command_cursor
 from mongomock import filtering
 from mongomock import helpers
+from mongomock import OperationFailure
 from sentinels import NOTHING
 from six import moves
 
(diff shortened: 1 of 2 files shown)


pcorpet (changes requested):
Please split your commits in two different Pull Requests.

Also can you update the tests to add the code that shows that the previous code was wrong please?

PR #645, opened 2020-07-19
Implement $switch operator

### LINKED ISSUE #631: Add '$switch' in the aggregate pipeline
When executing aggregate functions via MongoEngine making use of the aggregation pipeline operation of `$switch`, mongomock raise an `pymongo.errors.OperationFailure: Unrecognized expression '$switch'` exception.

Example
```
pipeline = [{'$group': {'_id': {'$switch': {'branches': [
            {'case': {'$gte': ['$COLUMN_VALUE', 50]}, 'then': 'HIGH'},
            {'case': {'$gte': ['$COLUMN_VALUE, 30]}, 'then': 'MEDIUM'}
        ],
            'default': 'LOW'
        }},
      "totalSum": {'$sum': "$AMOUNT"}}}]

MongoObjectModel.objects.all().aggregate(pipeline)
```

kosqx:
Fixes #631

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -629,8 +629,38 @@ class _Parser(object):
 
     def _handle_control_flow_operator(self, operator, values):
         if operator == '$switch':
-            branches = values['branches']
-            default = values.get('default')
+            if not isinstance(values, dict):
+                raise OperationFailure(
+                    '$switch requires an object as an argument, '
+                    'found: %s' % type(values)
+                )
+
+            branches = values.get('branches', [])
+            if not isinstance(branches, list):
+                raise OperationFailure(
+                    "$switch expected an array for 'branches', "
+                    'found: %s' % type(branches)
+                )
+            if len(branches) == 0:
+                raise OperationFailure(
+                    '$switch requires at least one branch.'
+                )
+
+            for branch in branches:
+                if not isinstance(branch, dict):
+                    raise OperationFailure(
+                        '$switch expected each branch to be an object, '
+                        'found: %s' % type(branch)
+                    )
+                if 'case' not in branch:
+                    raise OperationFailure(
+                        "$switch requires each branch have a 'case' expression"
+                    )
+                if 'then' not in branch:
+                    raise OperationFailure(
+                        "$switch requires each branch have a 'then' expression."
+                    )
+
             for branch in branches:
                 try:
                     case = self.parse(branch['case'])
@@ -638,12 +668,19 @@ class _Parser(object):
                     case = False
                 if case:
                     return self.parse(branch['then'])
-            return self.parse(default)
+
+            if 'default' not in values:
+                raise OperationFailure(
+                    '$switch could not find a matching branch for an input, '
+                    'and no default was specified.'
+                )
+            return self.parse(values['default'])
+
         # This should never happen: it is only a safe fallback if something went wrong.
         raise NotImplementedError(  # pragma: no cover
-            "Although '%s' is a valid conditional operator for the "
+            "Although '%s' is a valid control flow operator for the "
             'aggregation pipeline, it is currently not implemented '
-            ' in Mongomock.' % operator)
+            'in Mongomock.' % operator)
 
     def _handle_set_operator(self, operator, values):
         if operator == '$in':
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -216,6 +216,8 @@ class _Parser(object):
                 return self._handle_array_operator(k, v)
             if k in conditional_operators:
                 return self._handle_conditional_operator(k, v)
+            if k in control_flow_operators:
+                return self._handle_control_flow_operator(k, v)
             if k in set_operators:
                 return self._handle_set_operator(k, v)
             if k in string_operators:
@@ -225,7 +227,6 @@ class _Parser(object):
             if k in boolean_operators + \
                     text_search_operators + \
                     projection_operators + \
-                    control_flow_operators + \
                     object_operators:
                 raise NotImplementedError(
                     "'%s' is a valid operation but it is not supported by Mongomock yet." % k)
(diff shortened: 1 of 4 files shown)


mongomock/aggregate.py
```diff
@@ -625,6 +626,24 @@ def _handle_conditional_operator(self, operator, values):
...
+                try:
+                    case = self.parse(branch['case'])
+                except KeyError:
+                    case = False
+                if case:
+                    return self.parse(branch['then'])
+            return self.parse(default)
+        # This should never happen: it is only a safe fallback if something went wrong.
+        raise NotImplementedError(  # pragma: no cover
+            "Although '%s' is a valid conditional operator for the "
```
pcorpet: I think you forgot to update the text here: it's not a conditional operator but a control flow operator.

mongomock/aggregate.py
```diff
@@ -625,6 +626,24 @@ def _handle_conditional_operator(self, operator, values):
...
 
+    def _handle_control_flow_operator(self, operator, values):
+        if operator == '$switch':
+            branches = values['branches']
+            default = values.get('default')
+            for branch in branches:
+                try:
+                    case = self.parse(branch['case'])
+                except KeyError:
+                    case = False
```
pcorpet, line 668: Please add a test to cover that case (when the `case` key is missing).

pcorpet (changes requested):
Thanks for the implementation.

I have just  few comments before I merge.

mongomock/aggregate.py
kosqx: Looking at the [documentation](https://docs.mongodb.com/manual/reference/operator/aggregation/switch/#:~:text=must%20contain%20only%20a%20case%20field) the `case` is not optional:

> The objects in the branches array **must** contain only a `case` field and a `then` field.

Here I am trying to deal with the situation that the value does not exists, like this one already added to tests:
```py
{'case': '$missing_field', 'then': 'first case'},
```

I check the real MongoDB instance and the missing field in `case` jumps to next branch, but missing field in `then` skips entire `$switch`. 

I will extract `branch['case']` before `try-except` to show that code needs to handle only the exceptions raised by `self.parse(...)` and improve the tests.

mongomock/aggregate.py
```diff
@@ -626,6 +627,61 @@ def _handle_conditional_operator(self, operator, values):
...
+    def _handle_control_flow_operator(self, operator, values):
+        if operator == '$switch':
+            if not isinstance(values, dict):
+                raise OperationFailure(
+                    '$switch requires an object as an argument, '
+                    'found: %s' % type(values)
+                )
+
+            branches = values.get('branches', [])
+            if not isinstance(branches, list):
```
pcorpet: Please also accept tuples.

mongomock/aggregate.py
```diff
@@ -626,6 +627,61 @@ def _handle_conditional_operator(self, operator, values):
...
+                raise OperationFailure(
+                    "$switch expected an array for 'branches', "
+                    'found: %s' % type(branches)
+                )
+            if len(branches) == 0:
```
pcorpet: `if not branches` is equivalent and more pythonic

mongomock/aggregate.py
```diff
@@ -626,6 +627,61 @@ def _handle_conditional_operator(self, operator, values):
...
+
+            for branch in branches:
+                if not isinstance(branch, dict):
+                    raise OperationFailure(
+                        '$switch expected each branch to be an object, '
+                        'found: %s' % type(branch)
+                    )
+                if 'case' not in branch:
+                    raise OperationFailure(
+                        "$switch requires each branch have a 'case' expression"
```
pcorpet, line 657: missing a `to` before `have`, same below

tests/test__collection_api.py
```diff
@@ -3136,6 +3136,111 @@ def test__aggregate_project_if_null_expression(self):
...
+        ]
+
+        for switch_operator, expected_exception in tests_cases:
+            pipeline = [
+                {'$match': {'_id': 1}},
+                {'$project': {'result_field': switch_operator}},
+            ]
+            with self.assertRaises(mongomock.OperationFailure) as err:
+                self.db.collection.aggregate(pipeline)
+            self.assertEqual(expected_exception, str(err.exception))
```
pcorpet, line 3251: That (checking the exact exception string) seems like an overkill compared to other tests in here. Now that's it's here let's keep it, but if you do another change in mongomock, please only check the type of the error.

pcorpet (approved):
Great, thanks!

mongomock/aggregate.py
kosqx: I understand that omission of `to` in the message is ungrammatical.

I took all error messages from real connection to the MongoDB (version 4.2.6) via pymongo (version 3.10.1, the version 3.11.0 appends `full error:` info) using code similar to:

```py
try:
    collection.aggregate([
        {
            "$project": {
                'new_field': {
                    '$switch': {
                        'branches': [{}],
                    }
                }
            }
        },
    ])
except Exception as exc:
    print(str(exc))
```

This is also why periods on the end of messages are inconsistent:

```
$switch requires each branch have a 'case' expression
$switch requires each branch have a 'then' expression.
```

I am not sure what to do here. Should I:

- convert messages into grammatical sentences, or
- preserve real MongoDB+pymongo messages even if incorrect grammatically?

mongomock/aggregate.py
kosqx: Support for the `tuple` added. Validation of it has been added to the tests.

mongomock/aggregate.py
kosqx: Done, thanks!

mongomock/aggregate.py
pcorpet: Both are fine by me (either grammatically correct or like real MongoDB+pymongo). Having the exact same error message is not required because they tend to change it (here we could imagine that they would fix it at one point), so my take on that for now is: make sure to use the same exception type and use an error message that is understandable by a user. For the rest, I'm OK to go with the wind and sometimes be consistant with pymongo, sometimes make a better message.

Leave it as such so that we can get your PR in before the next release (that we should make soon as pymongo v3.11 is incompatible with the latest release of mongomock)

kosqx:
I am not familiar with the Pull Requests on GitHub. Is there anything left I need to do in order to have this PR merged? 

I will be grateful for any suggestions and help.

PR #646, opened 2020-07-25
Fix aggregation failing if index does not exists

dvvolynkin:


pcorpet (changes requested):
Thanks for the fix.

Please add a test in test__mongomock to ensure that there are no regression and that we have the same behavior than pymongo going forward.

dvvolynkin:
Done

mongomock/aggregate.py
```diff
@@ -315,7 +315,16 @@ def _handle_project_operator(self, operator, values):
         if operator == '$arrayElemAt':
             key, index = values
             array = self._parse_basic_expression(key)
-            return array[index]
+
+            if index < len(array):
+                return array[index]
```
pcorpet, line 376: What about negative values? Can you add a test for that as well please?

I guess the implementation could be easier and more pythonic with:
```py
try:
  return array[index]
except IndexError:
  raise KeyError('%s is not a valid index in array' % index)
```

mongomock/aggregate.py
```diff
@@ -315,7 +315,16 @@ def _handle_project_operator(self, operator, values):
...
+            # As I understood, this type of exception will be caught in the top
```
pcorpet: It's not slower: that's great, you can get rid of the comment

pcorpet (changes requested):

pcorpet (approved):

pcorpet:
Thanks Daniil

PR #649, opened 2020-07-31
Compatibility with version 3.11.0 of pymongo

### LINKED ISSUE #648: mongomock is not compatible with pymongo 3.11.0
The [just released](https://pypi.org/project/pymongo/#history) pymongo 3.11.0 introduced the `hint` parameter to the following methods:

- `add_update()`
- `add_replace()`
- `add_delete()`

This is causing the exceptions similar to:

```
TypeError: add_update() got an unexpected keyword argument 'hint'
```

kosqx:
The pymongo 3.11.0 introduced the `hint` parameter to the
`add_update()`, `add_replace()`, `add_delete()` methods

Fixes #648

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -352,18 +352,18 @@ class BulkOperationBuilder(object):
         self.insert(doc)
 
     def add_update(self, selector, doc, multi=False, upsert=False, collation=None,
-                   array_filters=None):
+                   array_filters=None, hint=None):
         if array_filters:
             raise NotImplementedError(
                 'Array filters are not implemented in mongomock yet.')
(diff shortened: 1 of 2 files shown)


pcorpet (changes requested):
Thanks a lot for fixing that as soon as you saw it. I was on it as well :-)

In general for new options I prefer to be over cautious and raise NotImplementedError when they are used until we find the right way to at least validate the option.

I've prepared a change that I'll try to submit today that does that.
https://github.com/mongomock/mongomock/compare/delete-hint

kosqx:
Thanks! Your fix is obviously superior. 

I wanted to just make tests pass. I noticed the problem when working on `$switch` and initially added a quickfix to check where is the problem.

PR #651, opened 2020-08-01
Handling truthy and falsy values the way MongoDB does it

### LINKED ISSUE #650: Falsy values in mongomock are different than in MongoDB
In MongoDB only the following values are falsy (the values that are considered false when encountered in a boolean context):

- `false`
- `null`
- `0`
- `undefined`

The official documentation points to [this description](https://docs.mongodb.com/manual/reference/operator/aggregation/and/#and-boolean-behavior):

> In addition to the `false` boolean value, $and evaluates as false the following: `null`, `0`, and `undefined` values. The $and evaluates all other values as true, including non-zero numeric values and arrays.

The Python (and in result also `mongomock`) have much longer list of falsy values. To name a few:

- `""` (empty string)
- `[]` (empty list)
- `{}` (empty dict)

Adding of the following test to the `tests/test__mongomock.py` will demonstrate the bug:

```py
    def test__aggregate_cond__falsy(self):
        self.cmp.compare_ignore_order.aggregate([
            {'$project': {
                'new_value': {'$cond': ['', 2, 3]},  # <-- only important line, notice the empty string
            }},
        ])
```

Right now this affects (probably only) `$cond` but in future will affect also `$switch`, `$anyElementTrue`, `$allElementsTrue`, `$and`, `$or`, `$not`.

kosqx:
Fixes #650

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -250,6 +250,14 @@ class _Parser(object):
                 if self._ignore_missing_keys:
                     yield None
 
+    def _parse_to_bool(self, expression):
+        """Parse a MongoDB expression and then convert it to bool"""
+        # handles converting `undefined` (in form of KeyError) to False
+        try:
+            return helpers.mongodb_to_bool(self.parse(expression))
+        except KeyError:
+            return False
+
     def _parse_basic_expression(self, expression):
         if isinstance(expression, six.string_types) and expression.startswith('$'):
             if expression.startswith('$$'):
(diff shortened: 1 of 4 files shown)


mongomock/aggregate.py
```diff
@@ -249,6 +249,14 @@ def parse_many(self, values):
                 if self._ignore_missing_keys:
                     yield None
 
+    def parse_to_bool(self, expression):
```
pcorpet: Please prefix your function name with `_` as it's supposed to be private.

pcorpet (approved):
All good. I'll wait for you to rebase and update the switch implementation as well.

mongomock/aggregate.py
kosqx: Done, thanks!

kosqx:
I consider this PR to be done and ready to be merged.

Issue #655, opened 2020-08-11
TTL index support - high level proposal


juannyG:
While using 
* mongomock==3.19.0
* mongonengine==0.20.0
* pymongo==3.10.1

I ran into an issue that appears with `update` that was resolved in https://github.com/mongomock/mongomock/pull/628

Sadly (for me) this was released in version 3.20.0 along side [a commit that raises a `NotImplementedError` when it sees TTL indexes](https://github.com/mongomock/mongomock/commit/cd034305d03534223db4b04d18252280fc29855e#diff-4901d650fcdfbe042f25fcaf722eb2cc) that makes my test suite explode with exceptions at this line when trying to upgrade from 3.19.0 to 3.20.0.

I see that [expiring data is listed in one of the missing features](
https://github.com/mongomock/mongomock/blob/develop/Missing_Features.rst) and before I went ahead and implemented anything, I thought it prudent to ask first if there were any considerations/gotchas/preferences of implementation - I haven't seen anything in the issues list, however.

I think the expiry feature could (should) be done in a two prong approach: 
1. [one for the TTL index ](https://docs.mongodb.com/manual/tutorial/expire-data/#expire-documents-after-a-specified-number-of-seconds)
2. [another for the specific clock time expiry](https://docs.mongodb.com/manual/tutorial/expire-data/#expire-documents-after-a-specified-number-of-seconds)

While (1) is being implemented, a similar `NotImplemented` approach could be used for (2) (i.e. if we see `expireAt` - sorry, `raise NotImplementedError`)

My initial reaction to solve (1): 
* adding a sort of specialization to the collections store of indexes to track `expireAfterSeconds` - I think this might be the easier part...
* when a document is first created, an extra hidden property is added to the "record", of "now + expiry" datetime, indicating when the record would/should be "deleted" - this feels a little trickier, mostly because I haven't dug into this area of the library (yet...:smile: )

The final piece of the puzzle lies in "OK - you've expired - get out." Per the mongo documentation:

> _The TTL feature relies on a background thread in mongod that reads the date-typed values in the index and removes expired documents from the collection._

Well - we don't have that - we're locked in the GIL (I wouldn't recommend async for this either). At some point during collection and/or document access, some kind of examination of the collection could occur, effectively executing the responsibility of the `mongod` background thread, and remove the expired records.

[...]

PR #656, opened 2020-08-13
Create a framework to silence some NotImplementedError.

pcorpet:
For #655

### DIFF
--- a/mongomock/__init__.py
+++ b/mongomock/__init__.py
@@ -88,7 +88,9 @@ __all__ = [
     'ObjectId',
     'OperationFailure',
     'WriteConcern',
+    'ignore_feature',
     'patch',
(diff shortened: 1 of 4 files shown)


hellocoldworld:
My tests suite broke on 3.20 because of the "TTL index" error and tested now using `mongomock.ignore_feature("TTL index")` and this LGTM, also I think it will bring high flexibility for the future of `mongomock`. Would you considerer publishing in pypi after merging?

juannyG:
:+1: Looks good.

non-senses:
Just to mention that as of Oct 19 the latest tag (3.20) does not have this patch. Looking forward to have it released in PiP as I'm forced to stick to 3.19 until then.

PR #658, opened 2020-08-13
Implement the $regexMatch operator.

### LINKED ISSUE #647: Support for '$regexMatch'
Hello, Using MongoMock version `3.19.0`, I was testing aggregation using new `$regexMatch` operator from mongoDB version `4.2`. Here are the docs: https://docs.mongodb.com/manual/reference/operator/aggregation/regexMatch/ .

It seems it's not supported yet. The following error arises when running the test:

```
  File "/Users/gcw/.pyenv/versions/3.7.6/envs/tqtq-api-env/lib/python3.7/site-packages/mongomock/aggregate.py", line 189, in parse
    raise OperationFailure("Unrecognized expression '%s'" % k)
pymongo.errors.OperationFailure: Unrecognized expression '$regexMatch'
```

This would be a feature request. Thanks.

pcorpet:
Fix #647

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -9,6 +9,7 @@ import itertools
 import math
 import numbers
 import random
+import re
 import warnings
 
 from sentinels import NOTHING
@@ -21,10 +22,12 @@ from mongomock import helpers
 from mongomock import OperationFailure
 
 try:
-    from bson import decimal128
+    from bson import Regex, decimal128
     decimal_support = True
+    _RE_TYPES = (helpers.RE_TYPE, Regex)
 except ImportError:
     decimal_support = False
+    _RE_TYPES = (helpers.RE_TYPE)
 
 _random = random.Random()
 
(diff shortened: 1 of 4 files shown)


mongomock/aggregate.py
```diff
@@ -384,6 +387,66 @@ def _handle_string_operator(self, operator, values):
...
+            for field in ('input', 'regex'):
+                if field not in values:
+                    raise OperationFailure("$regexMatch requires '%s' parameter" % field)
+            unknown_args = set(values) - {'input', 'regex', 'options'}
+            if unknown_args:
+                raise OperationFailure(
+                    '$regexMatch found an unknown argument: %s' % list(unknown_args)[0])
+
+            try:
+                input_value = self.parse(values['input'])
```
gwainer, line 419: Can `KeyError` be raised here and at line 410? The keys were checked at line 395.

gwainer (approved):
Thanks @pcorpet, looks good.

mongomock/aggregate.py
kosqx: The `self.parse()` is rising `KeyError` when given parameter references not existing field.

For example this will return `False` - just like real MongoDB:

```py
{"$regexMatch": {"input": "$missingField", regex: "foo"}}
```

gwainer: My bad, haven't checked the code for `self.parse()`. Thank you for the explanation @kosqx.

pcorpet: Thanks @kosqx for answering :-)

PR #659, opened 2020-08-14
Introduce Docker to ease local development

juannyG:
While starting local development on a feature, I found myself wrestling with certain dependencies that weren't explicitly clear to me (Mongo needs to be running for tests; what version? for example) - so I thought I might do some digging in the recent Travis builds to try and create a Docker image to help make my experience more portable - and also give me a better idea as to how Travis will behave when it sees my changes 😄 

The Dockerfile itself uses [pyenv](https://github.com/pyenv/pyenv) to prime all the python versions `tox` needs and also prepares the desired MongoDB installation (in this case 4.1.1).

This is a completely optional development route and should have no impact on any active development environments (Travis or otherwise).

The only "gotchas" I encountered have to do with the second and third commits:

1. I added `--cover-erase` to test commands [which removes coverage statistics from previous runs](
https://nose.readthedocs.io/en/latest/plugins/cover.html#cmdoption-cover-erase).  This was due to the `py34` run experiencing a UnicodeDecodeError from a previous run's `.coverage` artifact. This shouldn't effect Travis, as all the version executions are isolated from one another - so this should only affect local environments.
2. Pinning the `pylint` command to python3.6 to ensure the score remained a 10.00/10. python3.6 matches the python version Travis uses for the `pylint` command. For some reason, the container would use python3.4 by default, which, because of its deprecation notice, caused the score to go down to a 9.99/10

To use after cloning, one can simply enter the directory containing the `Dockerfile`, build, and run!
```
cd mongomock
docker build -t mongomock .
docker run --rm -i -t mongomock:latest tox
```

FYI: If this is deemed acceptable, I'll probably do a follow up PR with another secondary optional route to use `docker-compose` - this would require basically everything of what is here anyway, but moved around, for cleaner organization. This would allow for a docker only setup, or a docker-compose setup, providing even more flexibility.


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN pyenv install 2.7.18
+RUN pyenv install 3.4.10
+RUN pyenv install 3.5.9
+RUN pyenv install 3.6.11
+RUN pyenv install 3.7.8
+RUN pyenv install 3.8.5
+RUN pyenv install pypy2.7-7.1.1
+RUN pyenv local 2.7.18 3.4.10 3.5.9 3.6.11 3.7.8 3.8.5 pypy2.7-7.1.1
+
+ENV MONGODB=4.1.1
```
pcorpet: Please use a build arg here so that it's easier to test with a different version of MongoDB:
```
ARG MONGODB=4.1.1
ENV MONGODB=$MONGODB
```

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN eval "$(pyenv init -)" && eval "$(pyenv virtualenv-init -)"
+RUN pyenv install 2.7.18
+RUN pyenv install 3.4.10
+RUN pyenv install 3.5.9
+RUN pyenv install 3.6.11
+RUN pyenv install 3.7.8
+RUN pyenv install 3.8.5
+RUN pyenv install pypy2.7-7.1.1
+RUN pyenv local 2.7.18 3.4.10 3.5.9 3.6.11 3.7.8 3.8.5 pypy2.7-7.1.1
+
```
pcorpet, line 19: I suggest we add a comment here to tell people updating this file that they should keep it in sync with travis.yml's install step. Note that we should also add a comment in `travis.yml` to keep it in sync with this file.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN mongodb-linux-x86_64-$MONGODB/bin/mongod --version
+RUN mkdir -p /data/db
+RUN ln -s /mongodb-linux-x86_64-$MONGODB/bin/mongod /usr/bin/mongod
+
+RUN pip install tox
+
+ADD . mongomock
+WORKDIR mongomock
+RUN chmod u+x docker_entrypoint.sh
+ENTRYPOINT ["./docker_entrypoint.sh"]
```
pcorpet: Please add a newline the end of the file.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+ENV MONGODB=4.1.1
+RUN wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-$MONGODB.tgz
+RUN tar xzf mongodb-linux-x86_64-$MONGODB.tgz
+RUN mongodb-linux-x86_64-$MONGODB/bin/mongod --version
+RUN mkdir -p /data/db
+RUN ln -s /mongodb-linux-x86_64-$MONGODB/bin/mongod /usr/bin/mongod
+
+RUN pip install tox
+
+ADD . mongomock
```
pcorpet: I'd prefer to name the folder inside the container something different than mongomock: either `project` (like travis does) or `work` or something like this. Otherwise I think it might be confusing to have the folder `/mongomock/mongomock/`.

pcorpet: Note that adding . will end up adding a lot of things inside the container. I feel that it's ok when mounting it as a volume, but the copy here might be long. I suggest you cherry-pick strictly the files and folders that are required, or even drop the files entirely and force the use of volumes mounting.

In both cases, I suggest that you add a `.dockerignore` file to make sure the build step doesn't prepare all files in here to be copied (for instance the `.git` folder might be big).

pcorpet (changes requested):
Thanks a lot for this! I should admit I have a local Dockerfile doing half of what you're proposing and was thinking about committing it for a while.

I have a few comments inline, and have also some general comments:
- please add something to the README to document how to use Docker
- I think we could add a CMD in the Dockerfile to launch tox by default
- the documentation or the default CMD should allow easily to launch both tox for all envs as well as for only one env especially the default one (python 3.8.5 with pymongo and pyexecjs)
- add your name to the list of contributors if you're OK with that
- I'm OK for you to add the docker-compose.yml file now already (it makes mounting the local volume way easier)

Again, thanks a lot, this is super useful.

Dockerfile
juannyG: :100: to this

The volume mounting in vanilla Dockerfile is...less than appealing, for what I'd like to do. This is actually precisely why I was mentioning docker-compose originally - it's volume mounting is great and would remove the need to do a `docker build` every time you make a change.

However, I'm still going to hold off on the docker-compose changes for a few reasons:
* I want to make sure the vanilla Docker content is in a solid place
* Following that - some folks may not be using docker-compose, so I don't want to pigeon hole anyone into one way to do this

TL;DR - the `.dockerignore` or more granular `ADD`s  is a must have.

juannyG: Annnnd I think I've hit a snag with isolating this content from the container...

> `Exception: Versioning for this project requires either an sdist tarball, or access to an upstream git repository. `

Without the `.git` directory - tox fails on the `GLOB sdist-make: /mongomock/setup.py` with the above...unless I add it back.

My sketch for `docker-compose` wouldn't have this problem as:
* Dockerfile would only be lines 1-17 + `pip install tox`
* docker-compose.yml would handle mongo image to use
* docker-compose.yml can cleanly mount host folder to container FS...no copy/add/etc

juannyG:
Given what I found [here](https://github.com/mongomock/mongomock/pull/659#discussion_r472439376), I'm going to start on a side branch off this of the docker-compose version...this way I get it out of my head and something tangible for you to review/play with.

I see a couple of options:
* docker-compose is the way this stuff has to be run; no vanilla docker allowed
* support both, but allow the `.git` folder to get copied in
* a third option that I'm missing because I'm still new to this project and how to build it :smile: 

I'll wait to add documentation about this stuff once we land somewhere a littler firmer, but I believe I've addressed the rest of your comments.

pcorpet:
> docker-compose is the way this stuff has to be run; no vanilla docker allowed

Let's do that. We're trying to ease the local development and remove the need of installing a trillion stuff on the machine. Asking for docker-compose in addition to docker is really small and I'm totally for that option.

juannyG:
Alright - here we are in `docker-compose` land!

tests/test__bulk_operations.py
```diff
@@ -30,7 +30,7 @@ class BulkOperationsTest(TestCase):
     def setUp(self):
         super(BulkOperationsTest, self).setUp()
         if self.test_with_pymongo:
-            self.client = pymongo.MongoClient()
+            self.client = pymongo.MongoClient(host=os.environ.get('TEST_MONGO_HOST', 'localhost'))
```
juannyG, line 33: All of these host definitions are the cost of defining `mongo` as another service in docker-compose: it no longer lives on localhost in this setup.

Benefit: locally, we can define the version we want to use without having to worry about installation gotchas like path or tarball file structure changes.

.travis.yml
```diff
@@ -3,7 +3,7 @@ cache: pip
 os: linux
 dist: xenial
```
pcorpet: Please also add a comment about keeping this one in sync with the Dockerfile

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN pyenv install 3.5.9
+RUN pyenv install 3.6.11
+RUN pyenv install 3.7.8
+RUN pyenv install 3.8.5
+RUN pyenv install pypy2.7-7.1.1
+RUN pyenv local 2.7.18 3.4.10 3.5.9 3.6.11 3.7.8 3.8.5 pypy2.7-7.1.1
+
+RUN pip install tox
+
+CMD ["python"]
```
pcorpet: Is this command useful for you? Just making sure because my usage was always to run `tox -e py38-pymongo-pyexecjs`

README.rst
```diff
@@ -125,6 +125,31 @@ To download, setup and perfom tests, run the following commands on Mac / Linux:
...
+
+.. code-block:: bash
+
+ docker-compose run --rm mongomock tox -r
+
+Similarly, if you'd like to run tox against a specific environment in the container:
+
+.. code-block:: bash
+
+ docker-compose run --rm mongomock tox -e py35-pymongo-pyexecjs
```
pcorpet: Please use the latest and more complete version as documentation (py38-pymongo-pyexecjs)

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+    volumes:
+      - .:/project
+    depends_on:
+      - mongo
+
+  mongo:
+    # The image chosen here should be kept in sync with the .travis.yml mongo version
+    image: mongo:4.4.0
+    ports:
+      - "27017:27017"
```
pcorpet: I don't think you need to expose the port to the host: only the mongomock container should access it.

pcorpet (approved):
Thanks again for the great work.

Dockerfile
juannyG: So - we won't be able to do that here, because `py38-pymongo-pyexecjs` relies on mongo and that only comes up in docker-compose ... here in `Dockerfile` - mongo is not on the radar yet.

juannyG: That said - I usually just add something out of habit - we can remove it here, as `docker-compose` is the main driver and defines `command: tox` for us anyway! :smile:

juannyG:
My pleasure!

PR #660, opened 2020-08-18
TTL Indexes

juannyG:
Implement https://github.com/mongomock/mongomock/issues/655

Key points:

1. The first thing the `_iter_documents` method does now is remove expired documents. From what I could tell, this seemed to be the central point where any and all document examination occurred.
2. The TTL "time frame-of-reference" was made mock-able by using a centralized wrapper of `datetime.datetime.utcnow` - however further centralization can be handled separately. Given the stated desire, I wanted to try and plant this seed.

There are some other could improvements I can/would like to eventually add, specifically, some kind of `IndexWrapper` object to drastically improve the readability of the `_expire_documents` method, for example:

`idx.is_compound()` instead of `if len(index['key']) > 1`

but I wanted to get the core of the feature and where it might live in front of you first.

Feature and tests driven by: https://docs.mongodb.com/manual/core/index-ttl/

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1250,8 +1250,6 @@ class Collection(object):
         updater(doc, field_name, field_value)
 
     def _iter_documents(self, filter):
-        self._store.remove_expired_documents()
-
         # Validate the filter even if no documents can be returned.
         if self._store.is_empty:
             filter_applies(filter, {})
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1455,10 +1455,6 @@ class Collection(object):
     def create_index(self, key_or_list, cache_for=300, session=None, **kwargs):
         if session:
             raise_not_implemented('session', 'Mongomock does not handle sessions yet')
-        if 'expireAt' in kwargs:
-            raise_not_implemented(
-                'expireAt TTL index', 'Mongomock does not handle expireAt TTL index yet'
-            )
         index_list = helpers.create_index_list(key_or_list)
         is_unique = kwargs.pop('unique', False)
         is_sparse = kwargs.pop('sparse', False)
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1452,6 +1452,8 @@ class Collection(object):
     def create_index(self, key_or_list, cache_for=300, session=None, **kwargs):
         if session:
             raise NotImplementedError('Mongomock does not handle sessions yet')
+        if 'expireAt' in kwargs:
+            raise NotImplementedError('Mongomock does not handle expireAt TTL index yet')
         index_list = helpers.create_index_list(key_or_list)
         is_unique = kwargs.pop('unique', False)
         is_sparse = kwargs.pop('sparse', False)
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1462,7 +1462,7 @@ class Collection(object):
             index_dict['sparse'] = True
         if is_unique:
             index_dict['unique'] = True
-        if 'expireAfterSeconds' in kwargs:
+        if 'expireAfterSeconds' in kwargs and kwargs['expireAfterSeconds'] is not None:
             index_dict['expireAfterSeconds'] = kwargs.pop('expireAfterSeconds')
 
         existing_index = self._store.indexes.get(index_name)
@@ -1506,11 +1506,11 @@ class Collection(object):
                 raise TypeError(
                     '%s is not an instance of pymongo.operations.IndexModel' % index)
 
-        # TODO: expireAfterSeconds awareness(?)
         return [
             self.create_index(
                 index.document['key'].items(),
                 session=session,
+                expireAfterSeconds=index.document.get('expireAfterSeconds'),
                 unique=index.document.get('unique', False),
                 sparse=index.document.get('sparse', False),
                 name=index.document.get('name'))
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -6,7 +6,6 @@ try:
 except ImportError:
     from collections import Iterable, Mapping, MutableMapping
 import copy
-from datetime import datetime
 import functools
 import itertools
 import json
@@ -48,6 +47,7 @@ from six import string_types
 from six import text_type
 
 
+import mongomock
 from mongomock import aggregate
 from mongomock import ConfigurationError, DuplicateKeyError, BulkWriteError
 from mongomock.filtering import filter_applies
@@ -1500,7 +1500,7 @@ class Collection(object):
                         raise DuplicateKeyError('E11000 Duplicate Key Error', 11000)
                     indexed_list.append(index)
 
-        self._store.indexes[index_name] = index_dict
+        self._store.create_index(index_name, index_dict)
 
         return index_name
 
@@ -1529,7 +1529,7 @@ class Collection(object):
         else:
             name = index_or_name
         try:
-            del self._store.indexes[name]
+            self._store.drop_index(name)
         except KeyError:
             raise OperationFailure('index not found with name [%s]' % name)
 
@@ -2052,7 +2052,7 @@ def _current_date_updater(doc, field_name, value):
         if value == {'$type': 'timestamp'}:
             doc[field_name] = helpers.get_current_timestamp()
         else:
-            doc[field_name] = datetime.utcnow()
+            doc[field_name] = mongomock.utcnow()
 
 
 _updaters = {
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1244,6 +1244,8 @@ class Collection(object):
         updater(doc, field_name, field_value)
 
     def _iter_documents(self, filter):
+        self._store.remove_expired_documents()
+
         # Validate the filter even if no documents can be returned.
         if self._store.is_empty:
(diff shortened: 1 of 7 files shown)


mongomock/helpers.py
```diff
@@ -49,6 +49,10 @@ def dst(self, dt):
 DESCENDING = -1
 
 
+def utcnow():
```
pcorpet, line 52: Please document: what it does and why we handle it here

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
     def documents(self):
         for doc in six.itervalues(self._documents):
             yield doc
+
+    def remove_expired_documents(self):
```
pcorpet: I would prefer if the responsibility of never returning expired document would land in the store class. Therefore I believe this method should be protected and called in the other methods in this class.

Two reasons why I prefer it to be the responsibility of the store rather than the collection:
* the store is mocking the server's storage, whereas the collection is the client's view of it and the expiration of documents is done in the server
* there is no reasons ever that the store should returned expired document

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
...
+        for index in six.itervalues(self.indexes):
+            if index.get('expireAfterSeconds') is not None:
```
pcorpet: Nit - you could create a set of indexes that have an expireAfterSeconds field instead of checking them each time. I'm trying to optimize because this method is probably going to be called many many times.

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
...
+                self._expire_documents(index)
+
+    def _expire_documents(self, index):
+        # Ignore non-integer values
```
pcorpet, line 139: Please add a TODO to use a caching mechanism to avoid re-expiring the documents if we just did and that no document was added / updated.

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
...
+        except ValueError:
+            return
+
+        # Ignore commpound keys
+        if len(index['key']) > 1:
+            return
+
+        # "key" structure = list of (field name, direction) tuples
+        ttl_field_name = index['key'][0][0]
+        expired_ids = {
```
pcorpet: I'm not sure doc['_id'] have to be hashable, so I suggest you build a list instead of building a set.

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
...
+
+    def _value_meets_expiry(self, val, expiry):
+        val_to_compare = _get_dt_from_value(val)
+        try:
+            return (mongomock.utcnow() - val_to_compare).total_seconds() >= expiry
+        except TypeError:
+            return False
+
+
+def _get_dt_from_value(val):
```
pcorpet: please rename this function to `_get_min_datetime_from_value` or just `_get_min_datetime`

mongomock/helpers.py
pcorpet, line 52: Please update `_current_date_updater` to use this

mongomock/store.py
```diff
@@ -104,3 +107,51 @@ def __len__(self):
...
+            if self._value_meets_expiry(doc.get(ttl_field_name), expiry)
+        }
+
+        for exp_id in expired_ids:
+            del self[exp_id]
+
+    def _value_meets_expiry(self, val, expiry):
+        val_to_compare = _get_dt_from_value(val)
+        try:
+            return (mongomock.utcnow() - val_to_compare).total_seconds() >= expiry
```
pcorpet: Instead of calling utcnow on each document, I suggest you call utcnow only once in _expire_documents, or even in remove_expired_documents.

tests/test__collection_api.py
```diff
@@ -1313,9 +1323,90 @@ def test__create_index_wrong_type(self):
         with self.assertRaises(TypeError):
             self.db.collection.create_index([('value', 1, 'foo', 'bar')])
 
-    def test__create_ttl_index(self):
-        with self.assertRaises(NotImplementedError):
-            self.db.collection.create_index([('value', 1)], expireAfterSeconds=3600)
+    def test__ttl_index_record_expiration(self):
```
pcorpet: Please split in 3 test functions (as you can see, we are happy having multiple test functions, and here you might run into trouble because one test is leaking into the other, as your `drop` call suggests)

pcorpet (changes requested):
Wow, you made it! Please take a look at my comments.

mongomock/store.py
juannyG: Gotcha. I had thought of that originally, but once I saw that basically every method of the store would need to call this method, I started lurking elsewhere :wink:

mongomock/store.py
juannyG: I added some index helper methods to the store to ensure accurate state tracking while also satisfying this optimization.

juannyG:
Not sure what's up with the `pylint` run on that last build - locally I get 10.00/10

:thinking:

mongomock/collection.py
```diff
@@ -2050,7 +2052,7 @@ def _current_date_updater(doc, field_name, value):
         if value == {'$type': 'timestamp'}:
             doc[field_name] = helpers.get_current_timestamp()
```
pcorpet, line 2056: Note that we should also update this one somehow. Please add a TODO or find a way already.

mongomock/collection.py
```diff
@@ -2050,7 +2052,7 @@ def _current_date_updater(doc, field_name, value):
...
         else:
-            doc[field_name] = datetime.utcnow()
+            doc[field_name] = mongomock.utcnow()
```
pcorpet, line 2058: Please import it from helpers instead of mongomock

mongomock/store.py
```diff
@@ -78,29 +82,97 @@ def is_created(self):
...
+        except ValueError:
+            return
+
+        # Ignore commpound keys
+        if len(index['key']) > 1:
+            return
+
+        # "key" structure = list of (field name, direction) tuples
+        ttl_field_name = index['key'][0][0]
+        ttl_now = mongomock.utcnow()
```
pcorpet, line 151: Here as well, import it from helpers please.

mongomock/store.py
```diff
@@ -78,29 +82,97 @@ def is_created(self):
...
+        self._remove_expired_documents()
         return self._documents[key]
 
     def __setitem__(self, key, val):
         with lock:
             self._documents[key] = val
 
     def __delitem__(self, key):
-        del self._documents[key]
+        self._documents.pop(key)
```
pcorpet: why the change here? I find the `del` version more readable.

Also, shouldn't you need to remove expired documents before removing the key? I guess we are missing tests around deleting non existing documents.

pcorpet (approved):

mongomock/store.py
juannyG: I can do this but it has an interesting side-effect - [the test specific to mocking will fail](https://github.com/juannyG/mongomock/blob/ttl-index/tests/test__collection_api.py#L1389) because the "usage path" has changed.

There are two ways to go about this - the first I think will be less appealing, because I believe it goes against your original intent.

In `monogmock/store.py`

    from mongomock import helpers
    ...
    ttl_now = helpers.utcnow()

the test patch would then become

    with mock.patch('mongomock.store.helpers.utcnow') as mongomock_utcnow:

The implication of this is that now users would have to be aware of the different usages/locations in the library to target appropriately.

The alternative would allow for centralized mock that would affect all usages in the library that follow the same usage-pattern:

In `monogmock/store.py`

    import mongomock.helpers
    ...
    ttl_now = mongomock.helpers.utcnow()

the test patch would then become

    with mock.patch('mongomock.helpers.utcnow') as mongomock_utcnow:

From your earlier comments in the issue I originally opened, it seems as though the latter is the preferred approach: as long as anywhere in the library we want to allow mocking access uses `helpers.utcnow` in the same way - users would only need to mock in one place for "global" `utcnow` manipulation.

Subtle - but I wanted to raise this and get your input before I made any more changes. Thx!

mongomock/collection.py
juannyG: I'll try and revisit this separately

juannyG:
Separately - I was able to replicate the `pylint` errors travis is reporting - I needed to run `tox -r` to get a fresh tox env...but they don't seem related to my changes.

juannyG:
Digging deeper - using [six.raise_from](https://six.readthedocs.io/#six.raise_from) appears to clear up the pylint issues and appears backwards compatible (tests still pass).

LMK if you'd like for me to include those changes in this PR to keep the build nice and green :smile:

mongomock/store.py
pcorpet: I see your point, and you're entirely right. Let's just use `mongomock.utcnow` in our code as well. I'll try to make sure we don't use the `mongomock` import internally for other reasons. Maybe add a comment just above the `import mongmomock` explaining why we do it like that? Maybe just redirect readers to the README (no need for a huge explanation).

pcorpet:
I'll fix pylint errors at HEAD of develop.

juannyG:
Let me know if there is there anything else you'd like addressed or aren't sure about with these changes - thx!

PR #661, opened 2020-08-19
Implement the $split operator

khneal:
I didn't see any issues mentioning support for $split, but I figured a PR is better than filing an issue.

@pcorpet - Note the potential conflict with #658 in Missing_Features.rst. I don't mind fixing this PR if it's easier to merge that one first.

Let me know if any of the .gitignore entries are a problem.

Thanks!


khneal:
Side note - it was a bit strange to see all operators tested simultaneously `in test__aggregate_string_operations()`. Is there a reason for that, like making the tests run faster? If not, I would consider breaking those apart in a future PR.

.gitignore
```diff
@@ -3,6 +3,7 @@ ChangeLog
 
 #### Python
 *.py[co]
+/venv
```
pcorpet: Please drop the changes to this `.gitignore` file, and use a local gitignore instead.

pcorpet (changes requested):
Thanks a lot for the implementation.

Can you update `test__mongomock` also, to make sure the behavior stays the same between `pymongo` + `MongoDb` and `mongomock`?

Also, I believe you have no test where the split operation references a variable that does not exist:
```py
{'$split': ['$missingField', 'l']}
```
please add one.

.gitignore
khneal: Done. Good suggestion.

khneal:
> Can you update `test__mongomock` also, to make sure the behavior stays the same between `pymongo` + `MongoDb` and `mongomock`?

Done.

> Also, I believe you have no test where the split operation references a variable that does not exist:

Added. I almost committed one like that initially, but didn't see other string operations testing the same. I'm glad you pointed that out since it changed the implementation so the KeyError is caught, which in turn fixed a consistency issue I had in `test__mongomock`. Nice. 😃

pcorpet:
All good! Thanks for the quick turnaround. I'm going to merge now.

khneal:
Thank you! The PR process for this project was quick and easy, and kudos to you for all your efforts making it so.

I look forward to the next release. If there's a pre-release build available somewhere/somehow, I would be interested in pulling that into my workflow.

PR #663, opened 2020-08-26
Implement boolean operators

hellocoldworld:
Hello, added a new condition to the filter stage of my aggregation pipeline and suddendly my tests broke, because of the boolean operator.
I wanted to give it a try, but I'm afraid I'm being too simplistic maybe?
Or is this enough?

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -227,10 +227,9 @@ class _Parser(object):
                 return self._handle_string_operator(k, v)
             if k in type_convertion_operators:
                 return self._handle_type_convertion_operator(k, v)
-            if k in boolean_operators + \
-                    text_search_operators + \
-                    projection_operators + \
-                    object_operators:
+            if k in boolean_operators:
+                return self._handle_boolean_operator(k, v)
+            if k in text_search_operators + projection_operators + object_operators:
                 raise NotImplementedError(
                     "'%s' is a valid operation but it is not supported by Mongomock yet." % k)
             if k.startswith('$'):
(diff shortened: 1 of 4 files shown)


mongomock/aggregate.py
```diff
@@ -271,6 +270,19 @@ def _parse_basic_expression(self, expression):
             return helpers.get_value_by_dot(self._doc_dict, expression[1:], can_generate_array=True)
         return expression
 
+    def _handle_boolean_operator(self, operator, values):
+        if operator == '$and':
+            return all([self.parse(value) for value in values])
```
pcorpet: Please use `_parse_to_bool`, same below.

tests/test__collection_api.py
```diff
@@ -5611,6 +5607,56 @@ def test__aggregate_array_to_object(self):
             with self.assertRaises(mongomock.OperationFailure):
                 collection.aggregate(item)
 
+    def test_aggregate_project_with_boolean(self):
```
pcorpet, line 5610: Please also add tests in `test__mongomock` to make sure the behavior is consistent with pymongo + MongoDB

tests/test__collection_api.py
```diff
@@ -5611,6 +5607,56 @@ def test__aggregate_array_to_object(self):
...
+        collection = self.db.collection
+        collection.insert_many([{
+            'items': [],
+        }])
+
+        expect = [{'items': True}]
+        actual = self.db.collection.aggregate([
+            {'$project': {'_id': 0, 'items': {'$and': [True, True]}}}
```
pcorpet, line 5678: Please add more tests:
- one test using more than 2 items
- one test using less than 2 items (even 0)
- one test using something else than booleans

You don't need to cover all combination though, it's OK to have point checks only.

pcorpet (changes requested):
Thanks for adding this, it's exactly the right way.

I've added few comments in the tests, and I think you could also decide to cover extra cases where the syntax is not correct (for instance {'$and': 2}), but you can add a TODO and push those corner cases in a future PR.

Also make sure the CI passes, including the coverage check.

PR #674, opened 2020-10-20
performance improvement: create only one instance of _Filterer

tyarimi:
`_Filterer` is stateless, so no need to initialize it on every call to `filter_applies`. It cuts the filtering time by ±30%.

---
This change is [Reviewable](https://reviewable.io/reviews/mongomock/mongomock/674)

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -45,7 +45,7 @@ def filter_applies(search_filter, document):
     This function implements MongoDB's matching strategy over documents in the find() method
     and other related scenarios (like $elemMatch)
     """
-    return _Filterer().apply(search_filter, document)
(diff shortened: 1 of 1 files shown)


tyarimi:
@pcorpet can you please review and merge this PR? it improves performance dramatically.

pcorpet:
Thanks for the proposal. Note that the Filterer object was introduced for adding new features (that haven't made it yet to the main branch) and that I'll probably have to roll it back later on. As for now, if you're indeed seeing a +30% for performances, it's certainly worth it.

tyarimi:
@pcorpet thanks for merging the PR! I will try to find some time to introduce more performance improvements to the filtering logic, as we have thousands of tests depending on it.

PR #681, opened 2020-12-03
List  collection names with filter

hshmilov:
add filter support for list collection names 
https://pymongo.readthedocs.io/en/stable/api/pymongo/database.html#pymongo.database.Database.list_collection_names

i was missing the regex support :-)
also equal and none equal are supported

---
This change is [Reviewable](https://reviewable.io/reviews/mongomock/mongomock/681)

### DIFF
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -68,7 +68,7 @@ class Database(object):
         return self.list_collection_names(session=session)
 
     def _verify_list_collection_supported_op(self, keys):
-        if keys - list_collection_filter_allowed_operators:
+        if set(keys) - list_collection_filter_allowed_operators:
             raise NotImplementedError(
                 'list collection names filter operator {0} is not implemented yet in mongomock '
                 'allowed operators are {1}'.format(keys, list_collection_filter_allowed_operators))
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -18,7 +18,14 @@ try:
 except ImportError:
     _READ_PREFERENCE_PRIMARY = read_preferences.PRIMARY
 
-list_collection_filter_allowed_operators = frozenset(['$regex', '$eq', '$ne'])
+_LIST_COLLECTION_FILTER_ALLOWED_OPERATORS = frozenset(['$regex', '$eq', '$ne'])
+
+
+def _verify_list_collection_supported_op(keys):
+    if set(keys) - _LIST_COLLECTION_FILTER_ALLOWED_OPERATORS:
+        raise NotImplementedError(
+            'list collection names filter operator {0} is not implemented yet in mongomock '
+            'allowed operators are {1}'.format(keys, _LIST_COLLECTION_FILTER_ALLOWED_OPERATORS))
 
 
 class Database(object):
@@ -67,17 +74,11 @@ class Database(object):
 
         return self.list_collection_names(session=session)
 
-    def _verify_list_collection_supported_op(self, keys):
-        if set(keys) - list_collection_filter_allowed_operators:
-            raise NotImplementedError(
-                'list collection names filter operator {0} is not implemented yet in mongomock '
-                'allowed operators are {1}'.format(keys, list_collection_filter_allowed_operators))
-
     def list_collection_names(self, filter=None, session=None):
         """filter: only name field type with eq,ne or regex operator
 
         session: not supported
-        supported operator are $eq, $ne, $regex
+        for supported operator please see _LIST_COLLECTION_FILTER_ALLOWED_OPERATORS
         """
         field_name = 'name'
 
@@ -86,8 +87,8 @@ class Database(object):
 
         if filter:
             if not filter.get('name'):
-                raise NotImplementedError(' list collection {0} might be valid but is not '
-                                          ' implemented yet in mongomock'.format(filter))
+                raise NotImplementedError('list collection {0} might be valid but is not '
+                                          'implemented yet in mongomock'.format(filter))
 
             filter = {field_name: {'$eq': filter.get(field_name)}} \
                 if isinstance(filter.get(field_name), str) else filter
--- a/mongomock/database.py
+++ b/mongomock/database.py
@@ -66,10 +66,10 @@ class Database(object):
         return self.list_collection_names(session=session)
 
     def list_collection_names(self, filter=None, session=None):
-        """
-            filter: only name field type with eq,ne or regex operator
-            session: not supported
-            supported operator are $eq, $ne, $regex
+        """filter: only name field type with eq,ne or regex operator
+
+        session: not supported
+        supported operator are $eq, $ne, $regex
         """
 
         allowed_operators = ['$regex', '$eq', '$ne']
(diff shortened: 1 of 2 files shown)


mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
 
-    def list_collection_names(self, session=None):
+    def list_collection_names(self, filter=None, session=None):
+        """filter: only name field type with eq,ne or regex operator
+
+        session: not supported
+        supported operator are $eq, $ne, $regex
+        """
+
+        allowed_operators = ['$regex', '$eq', '$ne']
```
pcorpet: Please move this outside of the function (at top level module) and make it a frozenset.

mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
+        filed_name = 'name'
```
pcorpet: Did you mean field_name?

mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
+
+        def verify_supported_op(keys):
```
pcorpet: Move this outside of this function please, as it doesn't depend on anything that might change depending on the call.

mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
+            for key in keys:
```
pcorpet: Note that if you convert allowed_operators to a set, and make sure to use a set for keys, you can check directly `if keys - allowed_operators`.

mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
+                if key not in allowed_operators:
+                    raise NotImplementedError(
+                        'Mongomock list collection names filter'
```
pcorpet: Please mention which operator is not allowed but was used (keys - allowed_operators)

pcorpet: Please mention that the operator might be correct but has not been implemented in mongomock yet (the key is "yet" as we might implement it if many people asks for it).

mongomock/database.py
```diff
@@ -63,10 +65,39 @@ def collection_names(self, include_system_collections=True, session=None):
...
+                        ' operators are {0}'.format(allowed_operators))
+
         if session:
             raise NotImplementedError('Mongomock does not handle sessions yet')
 
+        if filter:
+            if not filter.get('name'):
+                raise NotImplementedError('Mongomock list collection names support name type only')
```
pcorpet: Again mention somewhere that the filter might be valid for pymongo but the operation is not implemented yet in mongomock.

tests/test__collection_api.py
```diff
@@ -5983,3 +5983,34 @@ def test_elem_match(self):
...
         self.assertEqual({1, 2}, ids)
+
+    def test_list_collection_names_filter(self):
+        self.db.create_collection('aggregator')
+        for day in range(10):
+            new_date = datetime.now() - timedelta(day)
+            self.db.create_collection('historical_{0}'.format(new_date.strftime('%Y_%m_%d')))
+
+        # test without filter
+        assert len(self.db.list_collection_names()) == 11
```
pcorpet: We use self.assertEqual in this test suite to have nicer error messages. Please do so in your new test

tests/test__collection_api.py
```diff
@@ -5983,3 +5983,34 @@ def test_elem_match(self):
         ids = set(doc['_id'] for doc in self.db.collection.find(
             {'arr': {'$elemMatch': {'$lt': 10, '$gt': 4}}}, {'_id': 1}))
         self.assertEqual({1, 2}, ids)
+
+    def test_list_collection_names_filter(self):
+        self.db.create_collection('aggregator')
+        for day in range(10):
+            new_date = datetime.now() - timedelta(day)
```
pcorpet: I suggest you compute datetime.now() only once at the top of your test: if the testing happens exactly at a date change the you would make sure that it doesn't impact the test.

pcorpet (changes requested):
Thanks a lot for adding the feature and taking the time to make all the checks pass.

mongomock/database.py
hshmilov: done

mongomock/database.py
hshmilov: typo fix

mongomock/database.py
hshmilov: nice

mongomock/database.py
hshmilov: done

mongomock/database.py
hshmilov: done

mongomock/database.py
hshmilov: done

mongomock/database.py
```diff
@@ -16,6 +18,8 @@
 except ImportError:
     _READ_PREFERENCE_PRIMARY = read_preferences.PRIMARY
 
+list_collection_filter_allowed_operators = frozenset(['$regex', '$eq', '$ne'])
```
pcorpet: Please name that _LIST_COLLECTION_FILTER_ALLOWED_OPERATORS (upper case for module constants, prefixed by `_` as it's supposed to be used in this module only).

mongomock/database.py
```diff
@@ -68,10 +72,37 @@ def collection_names(self, include_system_collections=True, session=None):
 
         return self.list_collection_names(session=session)
 
-    def list_collection_names(self, session=None):
+    def _verify_list_collection_supported_op(self, keys):
```
pcorpet: You are not using self, so you can move this as a module functionn (closer to the set above).

mongomock/database.py
```diff
@@ -68,10 +72,37 @@ def collection_names(self, include_system_collections=True, session=None):
...
+        if set(keys) - list_collection_filter_allowed_operators:
+            raise NotImplementedError(
+                'list collection names filter operator {0} is not implemented yet in mongomock '
+                'allowed operators are {1}'.format(keys, list_collection_filter_allowed_operators))
+
+    def list_collection_names(self, filter=None, session=None):
+        """filter: only name field type with eq,ne or regex operator
+
+        session: not supported
+        supported operator are $eq, $ne, $regex
```
pcorpet: I suggest we drop that and point to the frozenset instead, otherwise we would need to maintain the list of supported operators in two different places.

mongomock/database.py
```diff
@@ -68,10 +72,37 @@ def collection_names(self, include_system_collections=True, session=None):
...
+        """
+        field_name = 'name'
+
         if session:
             raise NotImplementedError('Mongomock does not handle sessions yet')
 
+        if filter:
+            if not filter.get('name'):
+                raise NotImplementedError(' list collection {0} might be valid but is not '
```
pcorpet: Please remove the white spaces before `list collection` and before `implemented`

pcorpet (approved):
Please finalize the last comments before I can merge your request.

mongomock/database.py
hshmilov: done

hshmilov:
can you please generate a minor release?

Issue #595, opened 2020-02-17
Implement '.' operator in the aggregation pipeline


ThibaultLemaire:
This is a feature request.

I get this error when trying to do a deep lookup:

```
NotImplementedError: Although '.' is valid in the 'localField' and 'as' parameters for the lookup stage of the aggregation pipeline, it is currently not implemented in Mongomock.
```

So I understand that this is a known issue, but I couldn't find an open issue here.

**When can we expect it to work (if ever) and/or what can I do in the meantime to work around that?**

---

For illustration, here's a pseudo query that triggers this error:

``` python
MyFirstModel.objects.aggregate({
    "$lookup": {
        "from": "my_third_models",
        "localField": "my_second_model.my_third_model_id", # This is the offending part
        "foreignField": "_id",
        "as": "my_third_model",
    }
})
```

ThibaultLemaire:
Okay, so I've found a workaround which translates roughly to the following using my example:

``` python
MyFirstModel.objects.aggregate(
    {
        "$lookup": {
            "from": "my_second_models",
            "localField": "my_second_model_id",
            "foreignField": "_id",
            "as": "my_second_model",
        }
    },
    {"$unwind": "$my_second_model"},
    {
        "$project": {
            "my_third_model": "$my_second_model.my_third_model_id", # This is supported
            # ... Other fields
        }
    },
    {
        "$lookup": {
            "from": "my_third_models",
            "localField": "my_third_model", # And now this works
            "foreignField": "_id",
            "as": "my_third_model",
        }
    },
)
```

The idea is to do each step of the lookup manually.

arunl:
I have initiated PR #686 to address this issue. If needed the original is available in my repo.

PR #686, opened 2020-12-31
Allow dot'ted field in connectFrom of $graphLookup aggregation pipeline

arunl:
This relates to Issue #595. This adds capability for using dot'ed parameters in connectionFrom field of $graph aggregation pipeline.

---
This change is [Reviewable](https://reviewable.io/reviews/mongomock/mongomock/686)


arunl:
Am pretty close to getting thru the pipeline. The code coverage is failing because my tests are not covering one additional statement, compared to the original. I could use help with which statement am I missing.  I am not able to get the details from codecove report. (It just could be that I connected my repo after this failure. Will give it a try after making some dummy changes.)

.gitignore
```diff
@@ -51,3 +51,4 @@ flycheck-*
 
 ### vim
 .ropeproject/
+*~
```
pcorpet: Please don't modify this file, instead update your local .gitignore to cover for those files.

mongomock/aggregate.py
```diff
@@ -952,7 +952,7 @@ def _handle_graph_lookup_stage(in_collection, database, options):
                 "Argument '%s' to $graphLookup must be string" % operator)
         if options[operator].startswith('$'):
             raise OperationFailure("FieldPath field names may not start with '$'")
-        if operator in ('connectFromField', 'as') and \
+        if operator in ('as') and \
```
pcorpet: Thanks for implementing, please simplify the condition here (`==` instead of `in`) and make it fit on one line.

Note that `operator in ('as')` would return true only for operator being`a` or `s`, not `as` (that might be the missing line in coverage).

mongomock/aggregate.py
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
...
+    def _recursive_get(match, nested_fields):
+        first_field = nested_fields[0]
+        head = match.get(first_field)
+        if len(nested_fields) == 1:
+            # final/last field reached
+            yield head
+        else:
+            # more to go, get remaining from list/dict
+            remaining_fields = nested_fields[1:]
+            if isinstance(head, list):
```
pcorpet: Please check for dict or tuples: `if isinstance(head, (list, tuple)):`

mongomock/aggregate.py
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
                 found_items.add(new_match['_id'])
         return new_matches
 
+    def _recursive_get(match, nested_fields):
+        first_field = nested_fields[0]
+        head = match.get(first_field)
+        if len(nested_fields) == 1:
+            # final/last field reached
+            yield head
+        else:
```
pcorpet: I suggest you return early in the branch above instead of adding the whole code in an `else` branch:

```py
if len(nested_fields) == 1:
  yield head
  return

# More to go: get remaining from list/dict.

File: mongomock/aggregate.py [hunk 5]
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
                 found_items.add(new_match['_id'])
         return new_matches
 
+    def _recursive_get(match, nested_fields):
+        first_field = nested_fields[0]
+        head = match.get(first_field)
+        if len(nested_fields) == 1:
+            # final/last field reached
```
pcorpet: Nit - Please start your comments with an upper case letter and end them with a dot.

mongomock/aggregate.py
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
                 found_items.add(new_match['_id'])
         return new_matches
 
+    def _recursive_get(match, nested_fields):
```
pcorpet: This function doesn't use anything apart from its parameters, so please move it at the module level.

mongomock/aggregate.py
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
...
+        first_field = nested_fields[0]
+        head = match.get(first_field)
+        if len(nested_fields) == 1:
+            # final/last field reached
+            yield head
+        else:
+            # more to go, get remaining from list/dict
+            remaining_fields = nested_fields[1:]
```
pcorpet: For better readability I suggest that you move this before line 991 and check remaining_fields instead:

```py
head = match.get(nested_fields[0])
remaining_fields = nested_fields[1:]
if not remaining_fields:
  yield head
```

mongomock/aggregate.py
```diff
@@ -984,6 +984,28 @@ def _find_matches_for_depth(query):
...
+                    # yield from _recursive_get(m, remaining_fields)
+                    for answer in _recursive_get(m, remaining_fields):
+                        yield answer
+            elif isinstance(head, dict):
+                # yield from _recursive_get(head, remaining_fields)
+                for answer in _recursive_get(head, remaining_fields):
+                    yield answer
+            else:
+                # field doesn't exist, just let go
+                pass
```
pcorpet: I'm not sure we need this code: dropping the else and pass statement should work the same.

tests/test__collection_api.py
```diff
@@ -3153,21 +3153,6 @@ def test__aggregate_graph_lookup_depth_filed_not_string(self):
             "Argument 'depthField' to $graphlookup must be a string",
             str(err.exception))
 
-    def test__aggregate_graph_lookup_dot_in_connect_from_field(self):
```
pcorpet, line 3156: Please add a test for dotted field in `as` if there isn't one (it should trigger a NotImplementedError)

pcorpet (changes requested):
Thanks a lot for implementing that feature.

I've noted that you're introducing a new architecture for test files. I would prefer that we do that in a separate PR. For now please try to fit in the existing test framework and maybe add a TODO to split it in smaller test modules.

Also I would very much like to add tests in test__mongomock that make sure that the behavior of mongomock stays the same as pymongo + MongoDB.

arunl:
@pcorpet  Re: changing test architecture.

Like to make sure I have understood you correctly. You are asking me to do the following:

1) Move the tests I have added into test__collection_api.py. These tests do not use live mongo server. 
2) Add more tests in test test__mongomock that compares the behavior of mongomock with pymongo + MongoDB.

Did I get you right?

arunl:
@pcorpet I've addressed all your comments, except one, which happens to be the most significant one, in my opinion. The pending task is to add tests in test__mongomock. 

I have reviewed how you have written other tests. I think it will be simple enough to translate the tests I have in "test__collections_api" into the dual comparison tests.  But it will likely take some time. I will need to switch context to my day job.

.gitignore
```diff
@@ -51,3 +51,4 @@ flycheck-*
 
 ### vim
 .ropeproject/
+
```
pcorpet: Please drop that line as well so that there is no change in this file for your PR.

pcorpet (changes requested):
Thanks for the changes. This seems OK to me. Just add the same tests in test__mongomock (without the expected results, as the expectations should just be to match the pymongo + MongoDB behavior).

arunl:
@pcorpet - I added tests in test__mongomock. Was very easy. You've set up the testing architecture very well.

tests/test__mongomock.py
```diff
@@ -3571,6 +3571,118 @@ def test_aggregate_project_with_boolean(self):
...
+                    'maxDepth': 2,
+                    'depthField': 'numConnections',
+                    'as': 'destinations'
+                }
+            }
+        ]
+
+        self.cmp_a.do.insert_many(data_a)
+        self.cmp_b.do.insert_many(data_b)
+        self.cmp_b.do.aggregate(query)
```
pcorpet: Please use self.cmp_b.compare, same in the other tests

pcorpet:
Thank you for the hard work. Feel free to send another PR to add your name in the contributors of the README (I should have suggested that earlier, sorry).

Issue #642, opened 2020-07-07
$mergeObjects support


miguelalvarezi:
First of all, thanks for this great library!

I'm trying to test an aggregation pipeline which includes a stage like the following:

```
{'$replaceRoot': {'newRoot': {'$mergeObjects': [default_doc, '$_id']}}}
```

I'm getting the following error:
 
```
pymongo.errors.OperationFailure: Unrecognized expression '$mergeObjects'
```

I suppose the library doesn't support `$mergeObjects` yet. Am I right?

Thanks!

pcorpet:
You're right it's not implemented. Although I think it would be easy to add if you want to send a PR.

I'm fixing the "bug" that it's raising an OperationFailure instead of a NotImplementedError.

PR #689, opened 2021-01-12
add support for mergeObjects operator

elbehery95:
As per https://github.com/mongomock/mongomock/issues/642

The current implementation follows the operator as defined such that on all null an empty doc is returned, and the docs are merged using ``dict.update`` method


pcorpet (changes requested):
Thanks for your PR.

Please add multiple tests in test__mongomock to ensure there are no regressions in the future.

elbehery95:
removed mergeobjects exception checking tests and added a single test__merge_objects under the aggregate tests

pcorpet:
Thanks, almost there. Please fix the remaining styling issues
https://travis-ci.org/github/mongomock/mongomock/jobs/754673610

pcorpet:
Thanks @elbehery95 for the hard work!

PR #690, opened 2021-01-12
fix use group operators in addFields

Navi6:
short PR fixing the use of grouping operators (min, max, sum and avg) used in addFields stage


pcorpet:
Great fix, thanks.

PR #692, opened 2021-01-18
Fix find data with projection is not immutable

duynguyenhoang:
When finding document with projection, currently data returned is mutable. It should be immutable.

How to reproduce:

```
# Prepare data
collection.insert_one({'_id': 1, 'years': {'last_job': 2010}})
item_before = collection.find_one({'_id': 1}, projection={'years': True})
# Try to update document dict, bug is here
item_before['years']['last_job'] = 2021

# Get item again
item_after = collection.find_one({'_id': 1})

# data should be 2010, but currently it is 2021
assert item_before['years']['last_job'] == 2010

```

Similar issue to old PR https://github.com/mongomock/mongomock/pull/643

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -234,7 +234,7 @@ def _project_by_spec(doc, combined_projection_spec, is_include, container):
 
     if not is_include:
         for key, val in iteritems(doc):
-            doc_copy[key] = val
+            doc_copy[key] = copy.deepcopy(val)
 
     for key, spec in iteritems(combined_projection_spec):
         if key == '$':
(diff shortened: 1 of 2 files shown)


pcorpet:
Thanks for your PR. I have two concerns with this change:
 - you write about immutable data after projection, however I feel this should be true even when not projecting (just finding data). If you agree then we should copy at the last moment, not when projecting.
 - there might be some big drawbacks in performances if we copy big objects. Did you do some stress tests?

duynguyenhoang:
- The implementation of finding data without copy is already immutable with `copy`, so we don't need to do copy at the last moment.
- I haven't do any stress test, but you are right deepcopy is bad in performances, let me use another lightweight of deepcopy instead.

PR #693, opened 2021-01-18
Add default and options to `MongoClient.get_default_database()`

leorochael:
Add the `default=` parameter to `MongoClient.get_default_database()` to equalize behavior with `pymongo`.

Forward remaining keyword arguments from `MongoClient.get_default_database()` to `MongoClient.get_database()`.

### DIFF
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -23,6 +23,7 @@ try:
     from bson.objectid import ObjectId
     import pymongo
     from pymongo import MongoClient as PymongoClient
+    from pymongo.read_preferences import ReadPreference
     _HAVE_PYMONGO = True
     _PYMONGO_VERSION = version.LooseVersion(pymongo.version)
 except ImportError:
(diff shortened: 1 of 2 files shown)


tests/test__mongomock.py
```diff
@@ -216,6 +217,31 @@ def client(uri):
         with self.assertRaises(ConfigurationError):
             c.get_default_database()
 
+    def test__getting_default_database_with_default_parameter(self):
+        def client(uri):
```
pcorpet: No need to create that here as you ony create one client.

Same below

pcorpet (approved):
Thanks for the fix.

tests/test__mongomock.py
leorochael: You're right. Sorry for the cargo-cult.

I've dropped them and rebased the PR.

leorochael:
I don't know what to make of [the one failing test](https://travis-ci.org/github/mongomock/mongomock/jobs/755692279).

I don't think it's related to the changes in this PR.

PR #628, opened 2020-05-22
Add support for read concern

abarto:
Adds support for setting the read concern at the collection level restoring compatibility with mongoengine 0.20.

Fixes: https://github.com/mongomock/mongomock/issues/625

PR #708, opened 2021-05-19
Support read_concern in database.get_collection()

m-aciek:
This argument is passed to underlying methods, so effectively is supported, only check removal is required.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -56,7 +56,6 @@ from mongomock import InvalidOperation
 from mongomock.not_implemented import raise_for_feature as raise_not_implemented
 from mongomock import ObjectId
 from mongomock import OperationFailure
-from mongomock.read_concern import ReadConcern
 from mongomock.results import BulkWriteResult
 from mongomock.results import DeleteResult
 from mongomock.results import InsertManyResult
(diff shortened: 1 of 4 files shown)


m-aciek:
Read concern is supported since 0.19.1 #628

m-aciek:
@pcorpet would it be possible for you to find time to review and merge this change?

pcorpet:
Hi @m-aciek and thanks for your contribution. If we're going to go that way, please make it possible to use a `with_options` (both in collection and database).

However two things to do before removing those checks:
- what should change in the behavior with a different read concern? From a quick glimpse to the doc, it seems that with a in-memory implementation with only one replica, all the read concerns will have the same effect. Please confirm.
- let's still make sure that setting a bad type as a read_concern param in any of those functions would return the same error that `pymongo` would raise.

m-aciek:
Thanks for looking into this.

Read concern works already in `with_options` of collection.

I think read concern shouldn't make a difference when using one replica, which we are mocking.

I wonder if I should bring a duck typing like in collection.with_options() (with using _WITH_OPTIONS_KWARGS) or just checking the type (first trying to import it from PyMongo, then from mongomock) is enough?

pcorpet:
You're right, I think we're fine like this.

Pease cover the new lines with some new tests, and I'll merge it.

m-aciek:
I've added missing test. I've added read concern type check support in Collection constructor (the constructor may be called in Database().get_collection() method). Also removed read concern type check from Collection().with_options() as it is now handled in the constructor.

m-aciek:
Pylint failure is not connected with changes here. I've added a new issue to track that: #709.

Is there a way to retrigger Codecov check?

m-aciek:
Added test for database.read_concern attribute, added setting default ReadConcern() if no value provided.

pcorpet:
Thanks for all the cleanups and tests.

PR #723, opened 2021-10-14
Support $objectToArray

eiszfuchs:
As I was tinkering with an aggregate pipeline I [found](https://stackoverflow.com/a/52601873) while looking for different ways of getting all properties within a collection, I noticed my tests were failing because `mongomock` doesn't support `$objectToArray` yet.

I tried to implement the function, but I wasn't 100% sure where to put it and how to properly test it.
So please, see this more of a draft or proposal. I will work on it further, but if you like my approach and can nudge me in the right direction, I'd be glad to contribute!

Oh, also, I didn't branch off the latest `develop` but of the last commit that was not failing, hope this is alright!

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -150,7 +150,8 @@ type_convertion_operators = [
     '$toInt',
     '$toDecimal',
     '$toLong',
-    '$arrayToObject'
+    '$arrayToObject',
+    '$objectToArray',
 ]
 
 
@@ -767,6 +768,23 @@ class _Parser(object):
                 'arrays used with $arrayToObject must contain documents '
                 'with k and v fields or two-element arrays'
             )
+
+        # Document: https://docs.mongodb.com/manual/reference/operator/aggregation/objectToArray/
+        if operator == '$objectToArray':
+            try:
+                parsed = self.parse(values)
+            except KeyError:
+                return None
+
+            if parsed is None:
+                return None
+
+            if not isinstance(parsed, (dict, collections.OrderedDict)):
+                raise OperationFailure(
+                    '$objectToArray requires an object input, found: {}'.format(type(parsed))
+                )
+
+            return [{'k': k, 'v': v} for k, v in parsed.items()]
 
         raise NotImplementedError(
             "Although '%s' is a valid type conversion operator for the "
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -6326,6 +6326,100 @@ class CollectionAPITest(TestCase):
             with self.assertRaises(mongomock.OperationFailure):
                 collection.aggregate(item)
 
+    def test_aggregate_object_to_array(self):
+        collection = self.db.collection
+
+        collection.insert_many([
+            {'items': None},
+            {'items': {'qty': 25}},
+            {'items': {
+                'size': {'len': 25, 'w': 10, 'uom': 'cm'},
+            }},
+        ])
+
+        expect = [{
+            'items': None,
+            'not_exists': None
+        }, {
+            'items': [
+                {'k': 'qty', 'v': 25},
+            ],
+            'not_exists': None
+        }, {
+            'items': [
+                {'k': 'size', 'v': {'len': 25, 'w': 10, 'uom': 'cm'}},
+            ],
+            'not_exists': None
+        }]
+
+        actual = collection.aggregate([
+            {'$project': {
+                'items': {'$objectToArray': '$items'},
+                'not_exists': {'$objectToArray': '$nothing'},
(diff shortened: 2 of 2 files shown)


pcorpet:
Starting the review. For reference the spec is [here](https://docs.mongodb.com/manual/reference/operator/aggregation/objectToArray/).

mongomock/aggregate.py
```diff
@@ -1446,6 +1462,7 @@ def _handle_match_stage(in_collection, database, options):
     '$lookup': _handle_lookup_stage,
     '$match': _handle_match_stage,
     '$merge': None,
+    '$objectToArray': _object_to_array_operation,
```
pcorpet: Please do not add it here: this is not a pipeline stage operator
https://docs.mongodb.com/manual/reference/operator/aggregation-pipeline/

mongomock/aggregate.py
```diff
@@ -190,10 +191,25 @@ def _merge_objects_operation(values):
...
+        })
+
+    return to_array
+
+
 _GROUPING_OPERATOR_MAP = {
     '$sum': _sum_operation,
     '$avg': _avg_operation,
     '$mergeObjects': _merge_objects_operation,
+    '$objectToArray': _object_to_array_operation,
```
pcorpet: Please do not add it here, it's not a group operator
https://docs.mongodb.com/manual/reference/operator/aggregation/group/#mongodb-pipeline-pipe.-group

mongomock/aggregate.py
```diff
@@ -190,10 +191,25 @@ def _merge_objects_operation(values):
     return merged_doc
 
 
+def _object_to_array_operation(values):
+    to_array = []
+
+    # Use predictable key order to be testable more easily
```
pcorpet: Please add a test and handle the case where the arg is not a dict

tests/test__collection_api.py
```diff
@@ -6481,6 +6481,53 @@ def test_aggregate_project_with_boolean(self):
...
+                {'k': '_id', 'v': 1},
+                {'k': 'values', 'v': {'foo': 'b', 'bar': 'e'}},
+            ]},
+            {'_id': 2, 'items': [
+                {'k': '_id', 'v': 2},
+                {'k': 'values', 'v': {'foo': 'c', 'bar': 'f'}},
+            ]},
+        ]
+        actual = collection.aggregate([{
+            '$project': {'items': {'$objectToArray': '$$ROOT'}}
```
pcorpet: This seems to be testing both the `$$ROOT` feature together with `$objectToArray`, maybe keep it simple and drop this first part of the test?

mongomock/aggregate.py
```diff
@@ -72,6 +72,7 @@
     '$stdDevPop',
     '$stdDevSamp',
     '$arrayElemAt',
+    '$objectToArray',
```
pcorpet: Please add it in the `type_conversion_operators` instead, then handle the implementation in  the `_handle_type_convertion_operator` method

pcorpet (changes requested):
Thanks for jumping in and contributing. Please address my comments.

eiszfuchs:
Thank you for your quick reply and review.
I might need a while to get back into making the requested changes, but your pointers are extremely helpful! I'll do my best!

tests/test__collection_api.py
eiszfuchs: I wasn't sure about it either. From where I'm standing, using `$objectToArray` only makes sense when using the pipeline values, but maybe there's another, simpler use-case for it. I will investigate!

(Because, when using pymongo as a driver, I can easily build the entries myself when using static values, there's no need to let MongoDB do that for me.)

pcorpet: The easiest to setup test is to check examples from the MongoDB doc: they're usually small and try not to use any other features than the ones required to demonstrate the concept.

I agree that you don't need MongoDB do that for you with static values, but that's what we want for a simple test: make sure it does something that you can do yourself (thus verify yourself).

pcorpet:
You can rebase, I've fixed the CI pipeline

eiszfuchs:
Wait, why was it closed? 

I rebased and changed my feature quite a lot, this time taking a lot of inspiration from the `$arrayToObject` operator, which I don't know how I could miss that the first time.

Let me know what you think!

pcorpet:
Thanks for the changes!

eiszfuchs:
@pcorpet Thanks for your time and help!

I'm seeing that there is still an issue with the order of the dicts in some Python versions (which breaks CI in `test_aggregate_object_to_array_with_example`); not sure why this didn't show up locally.

Issue #730, opened 2021-12-06
TypeError: if no direction is specified, key_or_list must be an instance of list with gridfs and pymongo 4.0


thequilo:
This issue came up while disscussing failing tests in the sacred package in IDSIA/sacred#852. The provided example works with pymongo <4.0 but fails with pymongo 4.0.

Contents of `test.py`:
```python
import mongomock

db = mongomock.MongoClient().db
from mongomock.gridfs import enable_gridfs_integration

enable_gridfs_integration()

import gridfs

s = gridfs.GridFS(db)
s.put(b'hello world')
```
Error message:
```
$ python test.py
Traceback (most recent call last):
  File "test.py", line 11, in <module>
    s.put(b'hello world')
  File ".../python3.8/site-packages/gridfs/__init__.py", line 135, in put
    grid_file.close()
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 320, in close
    self.__flush()
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 299, in __flush
    self.__flush_buffer()
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 291, in __flush_buffer
    self.__flush_data(self._buffer.getvalue())
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 272, in __flush_data
    self.__ensure_indexes()
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 215, in __ensure_indexes
    self.__create_index(self._coll.files, _F_INDEX, False)
  File ".../python3.8/site-packages/gridfs/grid_file.py", line 209, in __create_index
    collection.create_index(
  File ".../python3.8/site-packages/mongomock/collection.py", line 1467, in create_index
    index_list = helpers.create_index_list(key_or_list)
  File ".../python3.8/site-packages/mongomock/helpers.py", line 84, in create_index_list
    raise TypeError('if no direction is specified, '
TypeError: if no direction is specified, key_or_list must be an instance of list
```

oussjarrousse:
I have a good fix for this issue... But I am not sure how to contribute to this project...

oussjarrousse:
I hope this helps @thequilo !

oussjarrousse:
[...]

PR #731, opened 2021-12-09
adds if isinstance(key_or_list, ItemsView) to helpers.create_index_list(); fixing #730

oussjarrousse:
helpers.create_index_list() now check if the passed variable key_or_list is an instance of abc.ItemsView, if true it returns list(key_or_list). This update fixes the issue 730 https://github.com/mongomock/mongomock/issues/730 as is inspired by code from _index_list() in https://github.com/mongodb/mongo-python-driver/blob/v4.0/pymongo/helpers.py

### DIFF
--- a/mongomock/helpers.py
+++ b/mongomock/helpers.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict, abc
+from collections import OrderedDict
 from datetime import datetime, timedelta, tzinfo
 from distutils import version  # pylint: disable=no-name-in-module
 from mongomock import InvalidURI
@@ -7,6 +7,12 @@ from six.moves.urllib_parse import unquote_plus
 from six import PY3, iteritems, raise_from, string_types
 import time
 import warnings
+
+import sys
+if sys.version_info.major > 2:
+    from collections.abc import ItemsView
+else:
+    from collections import ItemsView
 
 
 # Get ObjectId from bson if available or import a crafted one. This is not used
@@ -86,7 +92,7 @@ def create_index_list(key_or_list, direction=None):
     """
     if isinstance(key_or_list, string_types):
         return [(key_or_list, direction or ASCENDING)]
-    if isinstance(key_or_list, abc.ItemsView):
+    if isinstance(key_or_list, ItemsView):
         return list(key_or_list)
     if not isinstance(key_or_list, (list, tuple)):
         raise TypeError('if no direction is specified, '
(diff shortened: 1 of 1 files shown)


mongomock/helpers.py
```diff
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, abc
```
pcorpet: Unfortunately we're still compatible with python2, so that won't fly. Either wait for pymongo to drop support for python2 or add a guard so that code doesn't fail in python2.

pcorpet (changes requested):
Thanks for your contribution.

Please add a test that fails without your change to avoid regression errors.

mongomock/helpers.py
oussjarrousse: Thank you for the review. Any suggestions or an example code on how to add a guard?

oussjarrousse: I have implemented a fix. I hope this time that will fly.

oussjarrousse:
I've implemented the required changes. looking forward to your review!

mongomock/helpers.py
```diff
@@ -8,6 +8,12 @@
 import time
 import warnings
 
+import sys
+if sys.version_info.major > 2:
+    from collections.abc import ItemsView
```
pcorpet, line 13: Please import `Iterable` instead

mongomock/helpers.py
```diff
@@ -86,6 +92,8 @@ def create_index_list(key_or_list, direction=None):
     """
     if isinstance(key_or_list, string_types):
         return [(key_or_list, direction or ASCENDING)]
+    if isinstance(key_or_list, ItemsView):
+        return list(key_or_list)
     if not isinstance(key_or_list, (list, tuple)):
```
pcorpet, line 97: Add it at the end here: `if not isinstance(key_or_list, (list, tuple, Iterable):` as we don't need to cast it as a list

mongomock/helpers.py
```diff
@@ -8,6 +8,12 @@
 import time
 import warnings
 
+import sys
```
pcorpet, line 11: Please move it as part of the import group above

pcorpet (changes requested):
I've started adapting mongomock to pymongo 4.0 and I've seen the error as well.

I will add pymongo 4 to the CI very soon so either I do it before you and I'll import your fix in my change, or you do it before without the need of adding a test despite what I asked earlier.

pcorpet:
As explained it got pulled in https://github.com/mongomock/mongomock/commit/151e12bda829e7a6d5361a6045b4e31d60120d80
so I'm going to close this PR.

oussjarrousse:
Whatever you see fit...

PR #735, opened 2022-01-04
Fix tests using pymongo

guludo:
By default, pip installs the latest version of pymongo with the previous tox configuration. As pymongo has already released version 4, which contains changes that breaks tests, let's instruct pip to install pymongo<4.0 until we have dealt with the incompatibilities.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,7 @@ deps=
     codecov
     coverage
     mock: mock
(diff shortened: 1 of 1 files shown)


kpk-pl (approved):

pcorpet:
Perfect, thanks!

PR #737, opened 2022-01-04
Improvements on set operators

guludo:
Hi there!

This PR does the following:
- Adds a helper function (`to_hashable()`) to allow us to better support equality checks and have document data in a hashable format. Such helper is used in the implementation of set operations;
- Adds an implementation for the `$setIntersection` operator;
- Updates `$setUnion` and `$setEquals` to use the same mechanism used by `$setIntersection`.

Also, there is an unrelated commit that fixes one of the errors raised by flake8.

Best regards,
Gustavo Sousa

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -1375,7 +1375,7 @@ def _handle_project_stage(in_collection, unused_database, options):
             except KeyError:
                 # Ignore missing key.
                 pass
-    if (method == 'include') == (include_id != False and include_id != 0):
+    if (method == 'include') == (include_id is not False and include_id != 0):
         filter_list.append('_id')
 
     if not filter_list:
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -878,44 +878,66 @@ class _Parser(object):
         if operator == '$in':
             expression, array = values
             return self.parse(expression) in self.parse(array)
-        if operator == '$setUnion':
-            result = []
-            for set_value in values:
-                for value in self.parse(set_value):
-                    if value not in result:
-                        result.append(value)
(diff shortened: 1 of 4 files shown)


pcorpet:
Hi @guludo and thanks a lot for all your recent contribution. I prefer to have small PRs that have only one goal.
You have identified several things in the same PR? great, create a PR for each, and I'll try to be quick at reviewing them.

Note that I squash each PR in a single commit before merging them on the `develop` branch.

guludo:
Hello, @pcorpet. Sorry about that. Well, I guess this PR could be split into 4:

1.  - `Use recommended way of comparing with False`

2.  - `Add to_hashable helper`
     - `Implement $setIntersection`
     
3. - `Use a more robust way of comparing results of set operations in tests`

4. - `Use a unified framework for some set operators`

It would also be possible to split (2) further, but I'm not sure that is worth it, since `Add to_hashable helper` would be seemingly useless until the subsequent PRs got merged.

I'd like to note that both (3) and (4) depend on (2). Should I create them as draft PRs until (2) is (hopefully) merged? Do you have a different preferred way of doing this?

pcorpet:
1. got merged as part of other changes :muscle: 

It seems that (3) would only affect tests, and I'm generally in favor in strengthening the tests before doing a change / refactoring. What do you think?

Otherwise to finalize your split:
- would to_hashable helper be useful in the current code (making the code more readable) or is it only used in $setIntersection? If it's useful, split it as a separate PR. If it's only used for the new feature, then you can package it as one PR.

pcorpet:
Hi @guludo , would you take the time to rebase and split this PR please?

PR #738, opened 2022-01-05
Implement $map operator

### LINKED ISSUE #677: Add support for $map
Hi,

Would it be possible to add support for $map operation in aggregations ?

I just added this operation to a database query and it breaks around 30 tests in my codebase.

Yours,

Laurent

guludo:
Hi there!

I believe this fixes #677.

Best regards,
Gustavo Sousa

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -78,7 +78,10 @@ project_operators = [
 control_flow_operators = [
     '$switch',
 ]
-projection_operators = ['$map', '$let', '$literal']
+projection_operators = [
+    '$let',
+    '$literal',
+]
 date_operators = [
     '$dateFromString',
     '$dateToString',
@@ -102,6 +105,7 @@ array_operators = [
     '$filter',
     '$indexOfArray',
     '$isArray',
+    '$map',
     '$range',
     '$reduce',
     '$reverseArray',
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -609,6 +609,43 @@ class _Parser(object):
 
             return None if None in parsed_list else list(itertools.chain.from_iterable(parsed_list))
 
+        if operator == '$map':
+            if not isinstance(value, dict):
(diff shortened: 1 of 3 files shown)


tests/test__collection_api.py
```diff
@@ -5601,6 +5601,77 @@ def test__aggregate_filter_wrong_options(self):
                 self.db.collection.aggregate(
                     [{'$project': {'filtered_items': {'$filter': option}}}])
 
+    def test__aggregate_map(self):
```
pcorpet, line 5692: Please add an equivalent test in test__mongomock

guludo: Done.

guludo:
Updated with an addition in `test__aggregate_map_errors()` to cover a missed line.

mongomock/aggregate.py
```diff
@@ -605,6 +609,43 @@ def _handle_array_operator(self, operator, value):
...
+            for k in value:
+                if k not in {'input', 'as', 'in'}:
+                    raise OperationFailure('Unrecognized parameter to $map: %s' % k)
+
+            input_array = self._parse_or_nothing(value['input'])
+
+            if input_array is None or input_array is NOTHING:
+                return None
+
+            if not isinstance(input_array, list):
```
pcorpet: Let's allow tuples as well

pcorpet (approved):

mongomock/aggregate.py
guludo: Got it. Done! Thanks.

Note: I think it would be interesting to have a way to "canonicalize" every input document so that we have a single expected type for each bson type. That `to_hashable` function implemented in #737 is something along those lines but with only "comparison" in mind.

pcorpet: Sure thing. Feel free to refactor and send a proposal as a separate PR. It's OK to start small and use the sanitizer only for one or two operators.

PR #743, opened 2022-01-11
Support 'number' as argument for filter $type

guludo:
Hi there.

This is a simple PR adding support for `number` as argument for filter operator `$type`, which is an alias grouping numeric types. 

For reference: https://docs.mongodb.com/manual/reference/operator/query/type/. It looks like this is supported by MongoDB from at least version 4.0 (haven't checked documentation for older versions).

Regards,  
Gustavo Sousa

### DIFF
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -1013,6 +1013,26 @@ class MongoClientCollectionTest(_CollectionComparisonTest):
         # self.cmp.compare.find_one(
         #     {'a': 1}, OrderedDict([('_id', 0), ('a', 0), ('b.c.e', 0), ('b.c', 0)]))
 
+    def test__find_type(self):
+        self.cmp.do.insert_many([
+            {'a': 1.2},  # double
+            {'a': 1},  # int
+            {'a': 1 << 32},  # long
+            {'a': decimal128.Decimal128('1.1')},  # decimal
+            {'a': 'a string value'},  # string
+            {'a': {'b': 1}},  # object
+            {'a': [1, 2, 3]},  # array or int
+            {'a': b'hello'},  # binData
+            {'a': ObjectId()},  # objectId
+            {'a': True},  # bool
+            {'a': datetime.datetime.now()},  # date
+        ])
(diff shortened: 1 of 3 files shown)


pcorpet (changes requested):
Very cool, thanks.

Please add a test also in test__mongomock to check that the behavior is the same with pymongo

guludo:
Hi @pcorpet !

As I was working in test__mongomock to add tests for `$type`, the tests showed that the current implementation already failed to reproduce mongo for other types, so I went ahead and added a commit to fix the behavior for them.

pcorpet:
thanks a lot for finding those. Would you be ok to split the fix in a different PR please? and have this PR only to introduce `number`. I like having a clean history and have only one commit per PR.

tests/test__mongomock.py
```diff
@@ -1013,6 +1013,26 @@ def test__find_projection_subdocument_lists(self):
...
+            {'a': decimal128.Decimal128('1.1')},  # decimal
+            {'a': 'a string value'},  # string
+            {'a': {'b': 1}},  # object
+            {'a': [1, 2, 3]},  # array or int
+            {'a': b'hello'},  # binData
+            {'a': ObjectId()},  # objectId
+            {'a': True},  # bool
+            {'a': datetime.datetime.now()},  # date
+        ])
+        for type_name, spec in mongomock.filtering.TYPE_MAP.items():
```
pcorpet, line 1030: I'd rather not rely on mongomock internals. Please list all the supported types here in a list.

guludo:
> thanks a lot for finding those. Would you be ok to split the fix in a different PR please?

Ok.

tests/test__mongomock.py
guludo: Okay.

Btw, is that the same for `test__empty_logical_operators`?

guludo:
I'm closing this one and will open separate pull requests.

tests/test__mongomock.py
pcorpet: You're right, I'll fix it.

Issue #748, opened 2022-01-27
Add support for $ifNull with multiple input expressions


David-Taub:
for reference: https://docs.mongodb.com/manual/reference/operator/aggregation/ifNull/

pcorpet:
Fixed by #749

PR #749, opened 2022-01-28
Added support for $ifNull multi field support, $isNumber, $isArray

David-Taub:
Added support for $ifnull with multiple input fields, added in MongoDB v4.4
solves issue #748

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -9,6 +9,7 @@ import functools
 import itertools
 import math
 import numbers
+from packaging import version
 import random
 import re
 import sys
@@ -16,6 +17,7 @@ import warnings
 
 from sentinels import NOTHING
 
+import mongomock
 from mongomock import command_cursor
 from mongomock import filtering
 from mongomock import helpers
@@ -878,13 +880,20 @@ class _Parser(object):
 
     def _handle_conditional_operator(self, operator, values):
         if operator == '$ifNull':
-            field, fallback = values
-            try:
-                out_value = self.parse(field)
-                if out_value is not None:
-                    return out_value
-            except KeyError:
-                pass
+            fields = values[:-1]
+            if len(fields) > 1 and version.parse(mongomock.SERVER_VERSION) <= version.parse('4.4'):
+                raise OperationFailure(
+                    '$ifNull supports only one input expression '
+                    ' in MongoDB v4.4 and lower'
+                )
+            fallback = values[-1]
+            for field in fields:
+                try:
+                    out_value = self.parse(field)
+                    if out_value is not None:
+                        return out_value
+                except KeyError:
+                    pass
             return self.parse(fallback)
         if operator == '$cond':
             if isinstance(values, list):
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -45,6 +45,7 @@ except ImportError:
 
 warnings.simplefilter('ignore', DeprecationWarning)
 IS_PYPY = platform.python_implementation() != 'CPython'
+SERVER_VERSION = version.parse(mongomock.SERVER_VERSION)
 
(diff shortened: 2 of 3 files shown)


tests/test__collection_api.py
```diff
@@ -3787,6 +3787,20 @@ def test__aggregate_project_if_null(self):
...
+        actual = self.db.collection.aggregate([
+            {'$match': {'_id': 1}},
+            {'$project': collections.OrderedDict([
+                ('_id', False),
+                ('a_and_b', {'$ifNull': ['$elem_a', '$elem_b', '<missing_both>']}),
+                ('b_and_a', {'$ifNull': ['$elem_b', '$elem_a', '<missing_both>']}),
+                ('b_and_c', {'$ifNull': ['$elem_b', '$elem_c', '<missing_both>']}),
+            ])}
+        ])
+        self.assertEqual([{'a_and_b': '<present_a>', 'b_and_a': '<present_a>',
```
pcorpet: Please break after `self.assertEqual(`

pcorpet (changes requested):
Thanks for the feature.

Please add a test in test__mongomock to make sure the feature stays aligned with pymongo + MongoDB behavior.

tests/test__collection_api.py
David-Taub: Note that I added a helper `get_mongodb_version` that checks the version of MongoDB
I added the newer $ifNull syntax that is supported only on versions higher than v4.4 which is currently used in this project. 
When this project upgrades into MongoDB v5, the test will pass and the syntax can be added

Do you think the PR can be approved if the aggregation fails depending on the MongoDB version

David-Taub:
@pcorpet, Note that I added a helper get_mongodb_version that checks the version of MongoDB
I added the newer $ifNull syntax that is supported only on versions higher than v4.4 which is currently used in this project.
When this project upgrades into MongoDB v5, the test will pass and the syntax will be used and not fail (added a check for MongoDB version for that)

tests/test__mongomock.py
```diff
@@ -3000,6 +3000,86 @@ def test__aggregate_count(self):
...
+        self.cmp.compare.aggregate([
+            {
+                '$project': {
+                    'a': {'$ifNull': ['$elem_a', '<missing_a>']},
+                    'b': {'$ifNull': ['$elem_b', '<missing_b>']},
+                }
+            }
+        ])
+
+        @skipIf(
```
pcorpet: Please deindent otherwise this function is nested in the other one.

mongomock/aggregate.py
```diff
@@ -156,6 +156,10 @@
     '$arrayToObject',
     '$objectToArray',
 ]
+type_operators = [
```
pcorpet: Please add those in a separate PR.

pcorpet (changes requested):
Thanks for your tests that revealed that we need more work.

I think that you're right that we could use a mechanism to handle different server versions in here, but I'd rather not introduce code that is not tested in the CI.

I think I'm going to first switch to MongoDB v5 for testing so that you can rebase and use that for the tests. Please remove the new helper and the skipIf parts.

pcorpet:
> Added support for $ifnull with multiple input fields, added in MongoDB v4.4 solves issue #748
Note that as opposed to what you wrote in your first comment here, the feature is only available *after* v4.4 not in 4.4.

pcorpet:
OK, this is done now, you have a `mongomock.SERVER_VERSION` that you can use

tests/test__mongomock.py
pcorpet: Done.

mongomock/aggregate.py
pcorpet: Done in #789

PR #753, opened 2022-02-02
Make find(projection=...) return copies of the values

### LINKED ISSUE #728: Changing find_one() result unexpectedly changes value stored in the database
Hi,

I've encountered a bug in Mongomock where changing the dict return value of `find_one()` changes the actual value that is stored in the database. This is different behaviour compared with MongoDB.

Here is a short test case:

```
import pymongo
import mongomock
import uuid
import sys

if len(sys.argv) == 2 and sys.argv[1] == 'mongodb':
    col = pymongo.MongoClient('localhost:27017')['testdb']['testcol']
else:
    col = mongomock.MongoClient()['mydb']['mycol']

# clear collection
col.delete_many({})

# insert test document
doc = {'foo': [{'bar': 'original value'}]}
col.insert_one(doc)

# find the document and update the returned dictionary
doc = col.find_one({}, {'foo': {'$elemMatch': {'bar': 'original value'}}})
doc['foo'][0]['bar'] = 'new value'

# fetch the document from the db again
doc = col.find_one({})

print(doc['foo'][0]['bar'])
```

```
$ python3 bug.py mongodb
original value

$ python3 bug.py mongomock
new value
```

I'm using version 3.23.0. Thanks.

niklasj:
Make sure that the values returned from find() and find_one() when a
projection is specified don't share any references to the actual db
documents.

Fixes #728

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1175,7 +1175,8 @@ class Collection(object):
                 doc_copy = self._copy_field(doc, container)
         else:
             doc_copy = _project_by_spec(
-                doc, _combine_projection_spec(fields),
+                self._copy_field(doc, container),
+                _combine_projection_spec(fields),
                 is_include=list(fields.values())[0],
                 container=container)
 
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -243,9 +243,23 @@ def _project_by_spec(doc, combined_projection_spec, is_include, container):
             elif isinstance(val, dict):
                 doc_copy[key] = _project_by_spec(val, spec, is_include, container)
         elif (is_include and spec is not NOTHING) or (not is_include and spec is NOTHING):
-            doc_copy[key] = copy.deepcopy(val)
+            doc_copy[key] = _copy_field(val, container)
 
     return doc_copy
+
+
+def _copy_field(obj, container):
+    if isinstance(obj, list):
+        new = []
+        for item in obj:
+            new.append(_copy_field(item, container))
+        return new
+    if isinstance(obj, dict):
+        new = container()
+        for key, value in obj.items():
+            new[key] = _copy_field(value, container)
+        return new
+    return copy.copy(obj)
 
 
 class BulkOperationBuilder(object):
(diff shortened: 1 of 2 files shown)


niklasj:
I'm not sure if this needs a test to compare it against the real MongoDB engine or not. But I'll add one if you would like it.

mongomock/collection.py
```diff
@@ -1181,7 +1181,8 @@ def _copy_only_fields(self, doc, fields, container):
                 doc_copy = self._copy_field(doc, container)
         else:
             doc_copy = _project_by_spec(
-                doc, _combine_projection_spec(fields),
+                self._copy_field(doc, container),
```
pcorpet: This does the job but in many cases some values will be copied several times. I suggest that you update the _combine_projection_spec instead and add a copy in the places where the are no copies yet (`doc_copy[key] = doc[key]` and `doc_copy[key] = val`)

pcorpet (changes requested):
Thanks for the PR. I don't think we need a test in test__mongomock for this one, so let's keep it like this.

mongomock/collection.py
niklasj: Ah… I think I see what you mean – if fields are excluded (e.g.`find_one(projection={'field': 0})`) or nested fields are used (e.g. `find_one(projection={'field.subfield': 1})`), then `_project_by_spec` would copy the values a second time.

I'll update the pull request accordingly, but I'm wondering a bit about the `container` parameter. As far as I can tell it's never anything but `dict`, so I'm wondering what its purpose is and if it would be OK to just `copy.deepcopy()` the values?

pcorpet: The `container` parameter was meant to use the implementation of the codec option `document_class`, if it's not too much trouble I'd like to keep it as such for now.

mongomock/collection.py
```diff
@@ -227,32 +227,23 @@ def _combine_projection_spec(projection_fields_spec):
...
-        else:
-            if is_include:
-                doc_copy[key] = doc[key]
-            else:
-                doc_copy.pop(key, None)
+                                 for sub_doc in val]
+            elif isinstance(val, dict):
+                doc_copy[key] = _project_by_spec(val, spec, is_include, container)
+        elif (is_include and spec is not NOTHING) or (not is_include and spec is NOTHING):
+            doc_copy[key] = copy.deepcopy(val)
```
pcorpet: Let's add at least a comment that this should use `container` if it's a dict-like object or may contain dict-like objects.

niklasj: I've changed it to use `_copy_field` instead now.

pcorpet:
Thanks or the change

PR #755, opened 2022-02-06
Fix behavior for $type filter on numeric types

guludo:
Hi there.

This is the first of 3 PRs made from splitting #743.

This makes changes so that we:

- Use correct type for "decimal"
- Ensure that we do not confuse a Python `bool` for an `int` (since
  `isinstance(True, int)` evaluates as True).
- Check on the bit-length of `int` values to check whether they should
  be considered "int" or "long" from Mongo's perspective.

This patch adds the missing test for $type in `test__mongomock.py`. Note
that the "array" type was purposefully ignored from `supported_types`,
as there is an issue that will be solved in a future patch:
`_type_op([1, 2, 3], t)` should return True for both `t = 'array'` and
`t = 'int'`.

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -21,6 +21,11 @@ except ImportError:
     DBRef = None
     _RE_TYPES = (RE_TYPE,)
 
+try:
+    from bson.decimal128 import Decimal128
+except ImportError:
+    Decimal128 = None
+
 _TOP_LEVEL_OPERATORS = {'$expr', '$text', '$where', '$jsonSchema'}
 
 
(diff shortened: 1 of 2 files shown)


guludo:
Force-pushed with changes required (comments where posted in #756).

pcorpet:
Please apply all the changes we added in #755 to simplify the map.

guludo:
> Please apply all the changes we added in #755 to simplify the map.

I'm not sure I follow. This is #755. Did you mean to squash #756 into this?

pcorpet:
Yes, sorry, I didn't mean to merge them, but all the remarks I made on #756 actually apply here.
Have `TYPE_MAP` be only about predicates.

guludo:
Hi @pcorpet,

Just updated this to have `TYPE_MAP` use only predicates. I preferred to use lambdas for all of them instead of adding a new function `_create_type_predicate`.

PR #756, opened 2022-02-06
 Fix behavior for $type filter on arrays

guludo:
This is the second PR (based on top #755) made from the split of #743.

Best regards,  
Gustavo Sousa

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -417,12 +417,16 @@ def _list_expand(f, negative=False):
     return func
 
 
-def _type_op(doc_val, search_val):
+def _type_op(doc_val, search_val, in_array=False):
     if search_val not in TYPE_MAP:
         raise OperationFailure('%r is not a valid $type' % search_val)
     elif TYPE_MAP[search_val] is None:
         raise NotImplementedError('%s is a valid $type but not implemented' % search_val)
-    return TYPE_MAP[search_val](doc_val)
+    if TYPE_MAP[search_val](doc_val):
+        return True
+    if isinstance(doc_val, (list, tuple)) and not in_array:
+        return any(_type_op(val, search_val, in_array=True) for val in doc_val)
+    return False
 
 
 def _combine_regex_options(search):
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -1046,6 +1046,7 @@ class MongoClientCollectionTest(_CollectionComparisonTest):
             'double',
             'string',
             'object',
+            'array',
             'binData',
             'objectId',
             'bool',
@@ -1058,6 +1059,7 @@ class MongoClientCollectionTest(_CollectionComparisonTest):
             {'a': 1.2},  # double
             {'a': 'a string value'},  # string
             {'a': {'b': 1}},  # object
+            {'a': [1, 2, 3]},  # array or int
             {'a': b'hello'},  # binData
             {'a': ObjectId()},  # objectId
             {'a': True},  # bool


mongomock/filtering.py
```diff
@@ -408,12 +413,27 @@ def func(doc_val, search_val):
...
-def _type_op(doc_val, search_val):
+def _type_op(doc_val, search_val, in_array=False):
     if search_val not in TYPE_MAP:
         raise OperationFailure('%r is not a valid $type' % search_val)
     elif TYPE_MAP[search_val] is None:
         raise NotImplementedError('%s is a valid $type but not implemented' % search_val)
-    return isinstance(doc_val, TYPE_MAP[search_val])
+
+    spec = TYPE_MAP[search_val]
+    if not isinstance(spec, tuple):
```
pcorpet: See my comments below to avoid this step

mongomock/filtering.py
```diff
@@ -408,12 +413,27 @@ def func(doc_val, search_val):
...
+        spec = (spec,)
+
+    for t in spec:
+        if callable(t):
```
pcorpet: In which case is it not callable?

mongomock/filtering.py
```diff
@@ -486,10 +506,14 @@ def operator_eq(doc_val, search_val):
...
+    'int': lambda v: (
+        isinstance(v, int) and not isinstance(v, bool) and v.bit_length() <= 32
+    ),
     'timestamp': None,
-    'long': (float,),
-    'decimal': (float,),
+    'long': lambda v: (
+        isinstance(v, int) and not isinstance(v, bool) and v.bit_length() > 32
+    ),
+    'decimal': Decimal128,
```
pcorpet: Please put Decimal128 in a tuple with one value so that this map only has values with `None` or a tuple.

If you want to cover for the case where Decimal128 is not defined, please solve it here directly `(Decimal128,) if Decimal128 else None`

mongomock/filtering.py
```diff
@@ -408,12 +413,27 @@ def func(doc_val, search_val):
     return func
 
 
-def _type_op(doc_val, search_val):
+def _type_op(doc_val, search_val, in_array=False):
```
pcorpet, line 420: Please do not add the `in_array` arg for now in here. Add it in the PR you need it.

pcorpet (changes requested):
Thanks for splitting in two PRs.

mongomock/filtering.py
guludo: Nice. It makes it explicit. Thanks.

mongomock/filtering.py
guludo: Oops... My bad.

mongomock/filtering.py
guludo: This is still used for lambdas. Unless we want use `'type_name': (lambda ...: ...,)` (which I personally do not find aesthetically appealing 🙂), in that case we could remove this.

mongomock/filtering.py
guludo: This is a result of my hasty splitting by just splitting lines without revising the content. In the next PR we also check for strings, this is why the check for callable here. I'll remove this first `if` here and add there.

mongomock/filtering.py
guludo: Hm... Actually, it seems it already is in the correct PR, isn't it?

guludo:
> Thanks for splitting in two PRs

There are actually 3 PRs: #755, #756 (this one), #757

mongomock/filtering.py
guludo: Done, I'll update the PR. (Note however that this change is actually part of #755 )

mongomock/filtering.py
guludo: Done. I'll update the PR (which is #755).

mongomock/filtering.py
pcorpet: Ah ok, got it. Please make sure to call it with `in_array=True` so that it's explicit below.

mongomock/filtering.py
pcorpet: OK. What do you think of having `TYPE_MAP` being a map with values being predicates instead of being either types or predicates:
```py
def _create_type_predicate(types):
  def _is_a(value):
     return isinstance(value, types)
```

Then:

```py
TYPE_MAP = {
  'double': _create_type_predicate(float),
  'string': _create_type_predicate(str),
  ...
  'null': None,
  ...
  'int': lambda v: ...,
}
```

mongomock/filtering.py
```diff
@@ -408,12 +413,27 @@ def func(doc_val, search_val):
...
+
+    for t in spec:
+        if callable(t):
+            if isinstance(t, type):
+                if isinstance(doc_val, t):
+                    return True
+            elif t(doc_val):
+                return True
+
+    if isinstance(doc_val, list) and not in_array:
```
pcorpet: please also check tuples: `if isinstance(doc_val, (list, tuple)) and not in_array:`

mongomock/filtering.py
guludo: That looks good. In that case, we could use lambdas directly instead of having a new function to create type predicates (e.g. `lambda v: isinstance(v, desired_type_or_types)` instead of  `_create_type_predicate(desired_type_or_types)`). The only problem I see with the sugested approach is with the next patch (in #757), which includes support for refering to other mongo types. For example, the "number" type is defined as in the diff below:

```diff
         isinstance(v, int) and not isinstance(v, bool) and v.bit_length() > 32
     ),
     'decimal': Decimal128,
+    'number': (
+        ('int', 'long', 'double')
+        + (('decimal',) if Decimal128 is not None else ())
+    ),
     'minKey': None,
     'maxKey': None,
 }
```
If we would only work with predicates, that diff above would probably have to be changed to the one below:

```diff
         isinstance(v, int) and not isinstance(v, bool) and v.bit_length() > 32
     ),
     'decimal': Decimal128,
+    'number': lambda v: any((
+        _type_op(v, 'int'),
+        _type_op(v, 'long'),
+        _type_op(v, 'double'),
+        _type_op(v, 'decimal') if Decimal else False,
+    )),
     'minKey': None,
     'maxKey': None,
 }
```

If that's okay for you, let's go for it!

mongomock/filtering.py
guludo: Done. Thanks!

mongomock/filtering.py
guludo: Done. Thanks!

mongomock/filtering.py
guludo: Done. Just updated #755.

guludo:
Just rebased this on top of #755.

pcorpet:
Simple and clean, please rebase so that I can merge easily

pcorpet (approved):

guludo:
Rebase done.

PR #757, opened 2022-02-06
Add support for "number" alias in $type

guludo:
This is the third (based on top #756) and final PR made from the split of #743.

Best regards,
Gustavo Sousa

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -508,6 +508,11 @@ TYPE_MAP = {
         isinstance(v, int) and not isinstance(v, bool) and v.bit_length() > 32
     ),
     'decimal': (lambda v: isinstance(v, Decimal128)) if Decimal128 else None,
+    'number': lambda v: (
+        # pylint: disable-next=isinstance-second-argument-not-valid-type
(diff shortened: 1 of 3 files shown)


guludo:
Just rebased this on top of #756

mongomock/filtering.py
```diff
@@ -508,6 +508,12 @@ def operator_eq(doc_val, search_val):
         isinstance(v, int) and not isinstance(v, bool) and v.bit_length() > 32
     ),
     'decimal': (lambda v: isinstance(v, Decimal128)) if Decimal128 else None,
+    'number': lambda v: any((
+        _type_op(v, 'int'),
```
pcorpet: Please use a simpler `isinstance(v, (int, float) + ((Decimal128,) if Decimal 128 else ()))`

guludo: Okay. Will do it. I'd like to notice though that I believe calling `_type_op()` here seems semantically more in tune with the meaning of the  definition of of type `number` by MongoDB. And less prone to errors too: for example, the simpler alternative proposed above is missing `and not isinstance(v, bool)`.

guludo: Done.

PR #761, opened 2022-02-17
Patch gridfs without using isinstance.

### LINKED ISSUE #758: enable_gridfs_integration broken for Python 3.10
Python 3.10 changed the way builtins are found at execution time (https://docs.python.org/3/whatsnew/3.10.html):

> Functions have a new __builtins__ attribute which is used to look for builtin symbols when a function is executed, instead of looking into __globals__['__builtins__']. The attribute is initialized from __globals__["__builtins__"] if it exists, else from the current builtins. (Contributed by Mark Shannon in [bpo-42990](https://bugs.python.org/issue42990).)

So, now the `isinstance` check in `GridFS.__init__` is not patched correctly. Example (only fails for Python 3.10):

```python
import mongomock
from mongomock.gridfs import enable_gridfs_integration
import gridfs

enable_gridfs_integration()
db = mongomock.MongoClient().db
fs = gridfs.GridFS(db)
```
raises:

```
  ...
  File ".../python3.10/site-packages/gridfs/__init__.py", line 74, in __init__
    raise TypeError("database must be an instance of Database")
TypeError: database must be an instance of Database
```

pcorpet:
Might fix #758

### DIFF
--- a/mongomock/gridfs.py
+++ b/mongomock/gridfs.py
@@ -1,75 +1,61 @@
+from unittest import mock
+
+from mongomock import Database as MongoMockDatabase, Collection as MongoMockCollection
+from mongomock.collection import Cursor as MongoMockCursor
+
+try:
+    from pymongo.collection import Collection as PyMongoCollection
+    from pymongo.database import Database as PyMongoDatabase
+    from gridfs.grid_file import GridOut as PyMongoGridOut, GridOutCursor as PyMongoGridOutCursor
+    _HAVE_PYMONGO = True
+except ImportError:
+    _HAVE_PYMONGO = False
+
+
+# This is a copy of GridOutCursor but with a different base. Note that we
+# need both classes as one might want to access both mongomock and real
+# MongoDb.
+class _MongoMockGridOutCursor(MongoMockCursor):
+
+    def __init__(self, collection, *args, **kwargs):
+        self.__root_collection = collection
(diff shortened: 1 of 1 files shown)


mongomock/gridfs.py
```diff
@@ -1,75 +1,61 @@
...
+
+
+# This is a copy of GridOutCursor but with a different base. Note that we
+# need both classes as one might want to access both mongomock and real
+# MongoDb.
+class _MongoMockGridOutCursor(MongoMockCursor):
+
+    def __init__(self, collection, *args, **kwargs):
+        self.__root_collection = collection
+        super(_MongoMockGridOutCursor, self).__init__(collection.files, *args, **kwargs)
```
thequilo, line 22: Just a question: Why do you pass arguments to `super`? The [docs](https://docs.python.org/3/library/functions.html#super) and the linked guide don't recommend this, but it doesn't hurt either (except for readability in my opinion)

thequilo (approved):
I don't know the internals of pymongo too well, so I can't tell if all details are right. My use-case with Python 3.10 works with this PR!

mongomock/gridfs.py
pcorpet: This is only historical python2-compatible code. I won't change it in this PR as I'm just moving those lines from below.

PR #768, opened 2022-04-22
Use class instead of function in patch

josefondrej:
Since `pymongo.MongoClient` is a class, we should also use a class to mock it.


pcorpet:
I'm unclear on the benefits, but it doesn't hurt :-)

josefondrej:
The benefit is that it enables mongomock to be used in tests for something like:
```
class MyCustomClient(pymongo.MongoClient):
    ...
```
Which was not possible previously. 
Not saying it is/is not a good idea to do something like ^^ :-)

PR #769, opened 2022-04-29
Improve field validation for documents

### LINKED ISSUE #720: bson.errors.invalidDocument: key * must not contains '.'
When using a real mongodb database I can include a entry like a below, but in mongomock it not possible.

{"object.name": value} 

When I add this in real database it works, but in mongomock doesn`t.

plaan:
Starting from mongodb 3.6 field names cannot contain the null character and top-field names cannot start with the dollar sign. Apart form those cases mongodb does permit storage of the dots and dollar signs ([mongodb v3.6](https://www.mongodb.com/docs/v3.6/reference/limits/#naming-restrictions)). Newer versions of mongodb add improved support ([mongodb v4.0](https://www.mongodb.com/docs/v4.0/reference/limits/#naming-restrictions), [mongodb v5.0](https://www.mongodb.com/docs/manual/reference/limits/)).

Fixes: https://github.com/mongomock/mongomock/issues/720

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -13,6 +13,7 @@ import warnings
 
 try:
     from bson import json_util, SON, BSON
+    from bson.errors import InvalidDocument
 except ImportError:
     json_utils = SON = BSON = None
 try:
@@ -252,6 +253,22 @@ def _copy_field(obj, container):
             new[key] = _copy_field(value, container)
         return new
     return copy.copy(obj)
+
+
+def _recursive_key_check_null_character(data):
+    for key, value in data.items():
+        if '\0' in key:
+            raise InvalidDocument(f'Field names cannot contain the null character (found: {key})')
+        if isinstance(value, Mapping):
+            _recursive_key_check_null_character(value)
+
+
+def _validate_data_fields(data):
+    _recursive_key_check_null_character(data)
+    for key in data.keys():
+        if key.startswith('$'):
+            raise InvalidDocument(f'Top-level field names cannot start with the "$" sign '
+                                  f'(found: {key})')
 
 
 class BulkOperationBuilder(object):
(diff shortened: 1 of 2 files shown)


mongomock/collection.py
```diff
@@ -254,6 +255,20 @@ def _copy_field(obj, container):
...
+    def recursive_none_key_check(data):
+        for key, value in data.items():
+            if "\0" in key:
+                raise InvalidDocument(f'Field names cannot contain the null character (found: {key})')
+            if isinstance(value, Mapping):
+                recursive_none_key_check(value)
+
+    recursive_none_key_check(data)
+    for key in data.keys():
+        if len(key) > 0 and key[0] == "$":
```
pcorpet: Use `key.startswith`

mongomock/collection.py
```diff
@@ -254,6 +255,20 @@ def _copy_field(obj, container):
     return copy.copy(obj)
 
 
+def _validate_data_fields(data):
+    def recursive_none_key_check(data):
```
pcorpet: This function isn't using any closure, so I'd rather not have it as a nested function

mongomock/collection.py
```diff
@@ -254,6 +255,20 @@ def _copy_field(obj, container):
...
+        for key, value in data.items():
+            if "\0" in key:
```
pcorpet: Please use single quotes for literal strings. `'\0'` (same in all your changes).

pcorpet (changes requested):
Thanks a lot for your contribution.

Please add a test that illustrates that recent versions accept weirder names.

mongomock/collection.py
plaan: Forgot about that one, thanks!

mongomock/collection.py
plaan: Moved it out of the function

mongomock/collection.py
plaan: Changed to using all single quotes

plaan:
Hi @pcorpet, wondering if you have the time for a second review. I made sure to add your suggested changes!

lluissalord:
Hi @pcorpet @plaan do we have any update on this? We are encountering with this issue, and we would like to solve it.

josemarcosrf:
This would be a great one so the mocking gets closer to the validation of the actual mongo. Any updates on getting this one merged?

pcorpet:
Thanks for pinging folks, it's working. I'll take a look now.

pcorpet (approved):

PR #772, opened 2022-05-03
dateTrunc support

MuriloScarpaSitonio:
Adds support for [$dateTrunc (aggregation)](https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateTrunc/).

I'm not sure if I should do some extensive testing, so please let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ flycheck-*
 
 ### vim
 .ropeproject/
+
+### pyenv
+.python-version
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -98,6 +98,7 @@ date_operators = [
     '$second',
     '$week',
     '$year',
+    '$dateTrunc',
 ]
 conditional_operators = ['$cond', '$ifNull']
 array_operators = [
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -51,3 +51,6 @@ flycheck-*
 
 ### vim
 .ropeproject/
+
```
pcorpet, line 54: Not related to this change, please revert

tests/test__mongomock.py
```diff
@@ -3956,6 +3956,20 @@ def test__aggregate_merge_objects(self):
         ]
         self.cmp.compare_ignore_order.aggregate(pipeline)
 
+    def test__aggregate__dateTrunc(self):
+        pipeline = [
+            {
+                "$group": {
```
pcorpet, line 3962: Use single quotes for string please.

mongomock/aggregate.py
```diff
@@ -539,6 +540,8 @@ def _handle_string_operator(self, operator, values):
 
     def _handle_date_operator(self, operator, values):
         out_value = self.parse(values)
+        if operator == '$dateTrunc':
+            return out_value['date']
```
pcorpet, line 544: Apparently this isn't doing much: please add some test cases, at least the examples here https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateTrunc/, and make sure to correct here.

pcorpet (changes requested):
Thanks a lot for your changes. Yes please add more testing. The logic I try to maintain is that a feature should not be half baked unless it raises an error when it's not functionally working the same way as MongoDb.

PR #793, opened 2022-07-14
Forward CodecOptions to BSON Encoder.

### LINKED ISSUE #745: Support UUID representation
As of `pymongo` v4+, it should error if used with default representation (Unspecified).

We could implement a first feature that accepts custom representation, and only fails (NotImplementedError) if writing/reading/comparing in different uuid representations for the same server. Otherwise if it's not Unspecified, we could handle UUID the same way we did before pymongo v4.

joshuashaffer:
Fixes #745.

This allows the use of UUID in pymongo4. 

Passes unit tests.

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -870,7 +870,7 @@ class Collection(object):
                             existing_document['_id'] = _id
                         if BSON:
                             # bson validation
-                            BSON.encode(document, check_keys=True)
+                            BSON.encode(document, check_keys=True, codec_options=self._codec_options)
                         existing_document.update(self._internalize_dict(document))
                         if existing_document['_id'] != _id:
                             raise OperationFailure(
@@ -1266,7 +1266,7 @@ class Collection(object):
             else:
                 return
         field_name = field_name_parts[-1]
-        updater(doc, field_name, field_value)
+        updater(doc, field_name, field_value, codec_options=self._codec_options)
 
     def _iter_documents(self, filter):
         # Validate the filter even if no documents can be returned.
@@ -2012,12 +2012,12 @@ class Cursor(object):
         return self
 
 
-def _set_updater(doc, field_name, value):
+def _set_updater(doc, field_name, value, codec_options = None):
     if isinstance(value, (tuple, list)):
         value = copy.deepcopy(value)
     if BSON:
         # bson validation
-        BSON.encode({field_name: value}, check_keys=True)
+        BSON.encode({field_name: value}, check_keys=True, codec_options=codec_options)
     if isinstance(doc, dict):
         doc[field_name] = value
     if isinstance(doc, list):
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -13,8 +13,10 @@ import warnings
 
 try:
     from bson import json_util, SON, BSON
+    from bson.codec_options import CodecOptions
 except ImportError:
     json_utils = SON = BSON = None
+    CodecOptions = type(None)
 try:
     import execjs
 except ImportError:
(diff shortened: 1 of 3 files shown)


joshuashaffer:
@pcorpet, does this seem to be an acceptable way of getting UUID support in pymongo4 ???

jschlyter:
Is setting `uuidRepresentation` via `MongoClient `, either via kwargs or via URI, also supported?

joshuashaffer:
> Is setting `uuidRepresentation` via `MongoClient `, either via kwargs or via URI, also supported?

Here's the test I used 

https://github.com/mongomock/mongomock/pull/793/files#diff-dec417cb2fe6e43e2fac60200a0dc7b290cbfd4bf3aa78a2505ae802b22c57fcR152

Via kwargs to the collection.

Greeengooo:
@pcorpet We are waiting for your opinion. Uuid representation is a great feature that will unblock many teams

pcorpet:
Thanks a lot for taking a look at this one.

I think the solution is on the right track, and thanks for adding the value in the BSON check. I'll comment in the review for smaller stuffs. The major point missing to me is comparing UUIDs in the same server from two different client objects using different UUID representations. I think the easiest way is to prevent any legacy Java / legacy C encoding as the python & standard encodings are the same. By prevent I mean raise a NotImplementedError.

mongomock/collection.py
```diff
@@ -13,8 +13,10 @@
 
 try:
     from bson import json_util, SON, BSON
+    from bson.codec_options import CodecOptions
 except ImportError:
     json_utils = SON = BSON = None
+    CodecOptions = type(None)
```
pcorpet, line 19: If it's possible, I'd rather have it to be `None`, and that you check that it's not None before using it.

mongomock/collection.py
```diff
@@ -74,6 +76,14 @@ class ReturnDocument(object):
 }
 
 
+def _bson_encode(document, codec_options):
+    if isinstance(codec_options, CodecOptions):
+        BSON.encode(document, check_keys=True,
```
pcorpet, line 81: Please break after `encode(`

pcorpet, line 81: Can you add a test where this fails because of the encoding?

tests/test__database_api.py
```diff
@@ -148,8 +149,10 @@ def test__with_options_pymongo(self):
             self.database.with_options(custom_document_class)
 
         custom_uuid_representation = codec_options.CodecOptions(uuid_representation=4)
-        with self.assertRaises(NotImplementedError):
-            self.database.with_options(custom_uuid_representation)
+        db = self.database
+        db.get_collection('yes_hello',
```
pcorpet, line 153: Please break after `get_collection(`

tests/test__database_api.py
```diff
@@ -148,8 +149,10 @@ def test__with_options_pymongo(self):
...
+                          codec_options=custom_uuid_representation).insert_one({'_id': uuid4()})
+        # db.yes_hello.insert_one({'_id': uuid4()})
```
pcorpet, line 155: Either drop this line or remove the `#`

pcorpet (changes requested):
Please add a test in `test__mongomock.py` to check that PyMongo + MongoDb and Mongomock have the same behaviors when using a `uuid_representation=4`

arturmkr:
@pcorpet 
@joshuashaffer 
Hello,
Do you need any help with this PR?

arturmkr:
Any updates on this?

hugo-azevedo:
Any news on when this fix is coming? Thanks

arturmkr:
PyMongo 4 was released one year ago.
But MongoMock still does not support.
@pcorpet , any plans on updating, supporting MongoMock?
Thanks!

mongomock/collection.py
ehiggs: Time to add ruff format lint stage to the CI? :)

LoicAvelin:
[Add custom type registry #911](https://github.com/mongomock/mongomock/pull/911)

Issue #706, opened 2021-04-24
$in operator not behaving like mongodb


ghost:
Hello,

$in operator behaves differently in mongomock than in mongodb.

For this document {'key': []}
Do a find with filter {'key': {'$in': [None, '', []]}} will not match the document in mongomock, but will match the document in mongodb.

Thanks for your help,

PR #795, opened 2022-07-16
fix in operator not behaving like mongodb when trying to compare to a…

orireiter:
… value of an empty list

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -265,7 +265,7 @@ def _iter_key_candidates_sublist(key, doc):
 
 
 def _force_list(v):
-    return v if isinstance(v, (list, tuple)) else [v]
+    return v if isinstance(v, (list, tuple)) and len(v) > 0 else [v]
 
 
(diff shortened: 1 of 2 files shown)


orireiter:
hi, i created a PR trying to fix the following issue:
fix #706 

would love to get feedback as to how to contribute the right way.
thanks!

mongomock/filtering.py
```diff
@@ -265,7 +265,7 @@ def _iter_key_candidates_sublist(key, doc):
 
 
 def _force_list(v):
-    return v if isinstance(v, (list, tuple)) else [v]
+    return v if isinstance(v, (list, tuple)) and len(v) > 0 else [v]
```
pcorpet: Please use the more pythonic boolean comparison `return v if v and isinstance(v, (list, tuple)) else [v]`

pcorpet (changes requested):
Thanks for your suggestion, please address my comments before I merge.

PR #797, opened 2022-08-09
Add partialFilterExpression parameter to index

Joe5K:
Parameter partialFilterExpression is [not added](https://github.com/mongomock/mongomock/blob/develop/mongomock/collection.py#L1540) to newly created index in mocked collection:
```python
def create_indexes(self, indexes, session=None):
        # ...

        return [
            self.create_index(
                index.document['key'].items(),
                session=session,
                expireAfterSeconds=index.document.get('expireAfterSeconds'),
                unique=index.document.get('unique', False),
                sparse=index.document.get('sparse', False),
                name=index.document.get('name'))
            for index in indexes
        ]
```
 which results in function [_ensure_uniques](https://github.com/mongomock/mongomock/blob/develop/mongomock/collection.py#L529) throwing an error when checking uniqueness:
 ```python
def _ensure_uniques(self, new_data):
        for index in self._store.indexes.values():
            # ...
            unique = index.get('key')
            is_sparse = index.get('sparse')
            partial_filter_expression = index.get('partialFilterExpression')
            # partial_filter_expression is always None
            # ...
            # find_kwargs = ...
            if partial_filter_expression is not None:
                find_kwargs = {'$and': [partial_filter_expression, find_kwargs]}
            answer_count = len(list(self._iter_documents(find_kwargs)))
            if answer_count > 1:
                # raised when two or more objects have the unique key parameter value = `null`
                raise DuplicateKeyError('E11000 Duplicate Key Error', 11000)  
```

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -1548,6 +1548,7 @@ class Collection(object):
                 index.document['key'].items(),
                 session=session,
                 expireAfterSeconds=index.document.get('expireAfterSeconds'),
+                partialFilterExpression=index.document.get('partialFilterExpression'),
                 unique=index.document.get('unique', False),
                 sparse=index.document.get('sparse', False),
                 name=index.document.get('name'))
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -1635,6 +1635,28 @@ class CollectionAPITest(TestCase):
 
         self.assertEqual(self.db.collection.count_documents({}), 4)
 
+    def test__ensure_partial_filter_expression_unique_indexes(self):
+        indexes = [
+            pymongo.operations.IndexModel([('obj.val_a', pymongo.ASCENDING), ('obj.val_b', pymongo.ASCENDING)],
+                                          unique=True,
+                                          partialFilterExpression={'obj': {"$exists": True}})
+        ]
+        self.db.collection.create_indexes(indexes)
+
+        # We should be able to add documents with duplicated `value` if `obj` doesn't exist
+        self.db.collection.insert_one({'value': 1})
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: mongomock/collection.py [hunk 1]
```diff
@@ -1548,6 +1548,7 @@ def create_indexes(self, indexes, session=None):
                 index.document['key'].items(),
                 session=session,
                 expireAfterSeconds=index.document.get('expireAfterSeconds'),
+                partialFilterExpression=index.document.get('partialFilterExpression'),
```
pcorpet, line 1551: Please add a test that would have failed without this line.

Joe5K: Test added.

tests/test__collection_api.py
```diff
@@ -1635,6 +1635,28 @@ def test__ensure_partial_filter_expression_unique_index(self):
...
+        indexes = [
+            pymongo.operations.IndexModel([('obj.val_a', pymongo.ASCENDING), ('obj.val_b', pymongo.ASCENDING)],
+                                          unique=True,
+                                          partialFilterExpression={'obj': {"$exists": True}})
+        ]
+        self.db.collection.create_indexes(indexes)
+
+        # We should be able to add documents with duplicated `value` if `obj` doesn't exist
+        self.db.collection.insert_one({'value': 1})
+        self.db.collection.insert_one({'value': 1})
```
Joe5K, line 1648: this is where exception occures without my bug fix

Joe5K:
can we merge this pls?

KING-SID:
@pcorpet saw that you're a reviewer on this, can we merge it in?

PR #815, opened 2022-11-07
Add $dateAdd operator

JPDevelop:
Add [$dateAdd](https://www.mongodb.com/docs/manual/reference/operator/aggregation/dateAdd/) date operator.

Does not support 'timezone' field.

Had to use python-dateutil for adding months to dates since it is relative and not an absolute value (e.g. 31/01 plus one month is 28/02)

### DIFF
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -9,6 +9,7 @@ import functools
 import itertools
 import math
 import numbers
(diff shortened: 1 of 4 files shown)


jamesfulford:
@JPDevelop @pcorpet is there anything I can do to help this merge? I'm going to work around not having this, but sure would love to remove my workaround.

PR #823, opened 2022-12-16
Forward codec options (#793 continuation)

ramnes:
Salut Pascal. :)

I'm trying to use mongomock as a test backend (through [Mongo-Thingy](https://github.com/Refty/mongo-thingy) and [mongomock-motor](https://github.com/michaelkryukov/mongomock_motor)), but the lack of UUID support makes it impossible.

So here is a continuation PR of #793, so that #745 can get closed; it seems like there are quite a few other persons interested in UUID support at this point (cc @Kalbi @elanou @joshuashaffer @jschlyter @Greeengooo @arturmkr @hugo-azevedo).

Looking at your GitHub activity, I seem to understand that you're busy with other things than mongomock at the moment, so I'll do my best to answer ASAP to any kind of feedback.

Thanks for the great library!

### DIFF
--- a/mongomock/codec_options.py
+++ b/mongomock/codec_options.py
@@ -76,6 +76,13 @@ class CodecOptions(collections.namedtuple('CodecOptions', _FIELDS)):
         opts.update(kwargs)
         return CodecOptions(**opts)
 
+    def to_pymongo(self):
+        if not codec_options:
+            return None
+        return codec_options.CodecOptions(
+            uuid_representation=self.uuid_representation,
+            unicode_decode_error_handler=self.unicode_decode_error_handler)
+
 
 def is_supported(custom_codec_options):
 
--- a/mongomock/codec_options.py
+++ b/mongomock/codec_options.py
@@ -50,8 +50,8 @@ class CodecOptions(collections.namedtuple('CodecOptions', _FIELDS)):
 
         if uuid_representation is None:
             uuid_representation = _DEFAULT_UUID_REPRESENTATION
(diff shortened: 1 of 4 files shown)


lewis-kori:
@ramnes were you able to proceed without the patches?
We're also experiencing the same issues.

ramnes:
I don't understand your question. Most of my test suite passes with these patches applied; a few don't but that's not related to UUID support.

lewis-kori:
@ramnes  Whenever I run my test suite, they fail because the codecs aren't forwarded to the BSON encoder and the [UUID validation check here](https://github.com/mongodb/mongo-python-driver/blob/master/bson/binary.py#L266-L274) fails.
So I was wondering how you were able to bypass these two issues.
I'm using mongomock v4.1.2

with mongoengine which has a dependency on pymongo v4.3.3.

ramnes:
By using this patch?

ramnes:
I'm closing this PR since I've lost interest in this. I'll leave the branch in my repository, feel free to rebase any future PR on it if you're willing to take up the torch. :)

lewis-kori:
thanks, @ramnes 
I haven't used the patch yet. seems like you have.
are you able to install it directly with pip or GitHub in a CI/CD pipeline?
we do need this working in such a workflow, but we aren't strict on deadlines.

ramnes:
https://adamj.eu/tech/2019/03/11/pip-install-from-a-git-repository/

lewis-kori:
thank you!

PR #864, opened 2023-11-10
Begin $convert implementation

MrSnapperVibes:
Does not include `onError`, `onNull`, or any `to` string identifier that has not already been implemented.


MrSnapperVibes:
Before I continue work on this issue, I would like to know if you have any feedback or suggestions for me. I want to make sure that my code is aligned with your standards and vision, and that it will be useful for the project. Thank you for your time and attention. 🙂

Issue #350, opened 2017-09-11
Why Not add a data persistence level into mongomock?


MacHu-GWU:
Since mongomock use ``Document. _documents`` to hold the data, why can't add methods ``Database.load(path, **options)`` and ``Database.dump(path, **options)``?

pcorpet:
Hi @MacHu-GWU , sorry for the long delay in the answer. I think we could probably add some persistence level in mongomock. However would we do that, I think we should not augment the Database API so that the database API stays as close as possible as the `pymongo` one.

So what we could have would be top module functions that would dump/load all indices and documents from a DB. Note that they could actually be in a separate package (no need to be inside mongomock) and work both on mock and real DB. See for instance this [answer on SO](https://stackoverflow.com/questions/24610484/pymongo-mongoengine-equivalent-of-mongodump). What do you think?

nguyenhn-systran:
Hi @pcorpet , could you tell me whether MongoMock can now load data from a file ?
I'd like to see whether I can use MongoMock in the integration testing, using test data coming from the production dump. 
Thanks for information.

pcorpet:
Sorry, there's no one-liner to do that however as explained above it should be easy to implement.

fopina:
Adding persistence to mongomock would allow to use pymongo/mongoengine for local development, without actually running a mongo instance, :+1: on that :)

wangwalton:
Hi ! I'm happy to make contributions on adding persistance to MongoMock. Can anyone provide some basic overview and pointers on getting started? @pcorpet

PR #881, opened 2024-02-19
Optional file storage for data persistence

ikondov:
The persistence topic has been already addressed in issues #350  and #406. We have another use case where the tests of the command line interface require saving the database and making it available across several processes. Therefore, some temporary file storage is necessary.

This can be accomplished by a small extension of the classes in `store.py`. The file store is not intended to keep database state instantly on disk. Rather the database is eventually dumped to a file just before the `ServerStore` object is destroyed.

~~It is a work in progress: TODO: a transparent method to pass the option to use a database file with a database file path. This can be accomplished e.g. by using an environment variable.~~

The persistence is activated either by setting the environment variable MONGOMOCK_SERVERSTORE_FILE to a non-empty JSON file (initialized with '{}'):

```bash
export MONGOMOCK_SERVERSTORE_FILE=/full/path/to/mongomock_file.json
```

or create a custom `ServerStore` object by passing the `filename` keyword argument:

```python
from mongomock.store import ServerStore
from mongomock import MongoClient

mongo_client = MongoClient(_store=ServerStore(filename='/full/path/to/mongomock_file.json'))
```


ikondov:
@pcorpet: Addressing your suggestion in #350, the Database API is not modified at all. Only the storage classes are extended with serialization methods and the `ServerStore` class with a file dump method. The methods to activate persistence are tested in the test module (see [test__persistence.py](https://github.com/mongomock/mongomock/pull/881/files#diff-ee44f7ea1e85a9d9633a4e64f37e3a5f3f6f3429955fae5e7a837ac8c9f3d207)). By default it is deactivated.

Issue #759, opened 2022-02-11
Support `.$id` and other fields of `DBRef` objects.


cypherlou:
Hi MongoMock team.

I've been using your brilliant library for a while and have recently encountered an issue when producing test cases for [Beanie with Linked documents](https://roman-right.github.io/beanie/tutorial/relations/).

```txt
mongomock==3.23.0
mongomock-motor==0.0.2
beanie==1.8.12
```

When attempting to create a linked document I get the following error;
```python
    def to_ref(self):
        if self.id is None:
            raise DocumentWasNotSaved("Can not create dbref without id")
>       return DBRef(self.get_motor_collection().name, self.id)
E       AttributeError: 'AsyncIOMotorCollection' object has no attribute 'name'

.env/lib/python3.8/site-packages/beanie/odm/documents.py:1174: AttributeError
```

I have resolved the problem with the following change to `beanie/odm/documents.py`;
```diff
1094a1095
>         collection_meta.motor_collection.name = collection_meta.name
```

More explicitly, adding the collection name in a `name` attribute of `collection_meta.motor_collection`.

```python
    @classmethod
    def get_motor_collection(cls) -> AsyncIOMotorCollection:
        """
        Get Motor Collection to access low level control

        :return: AsyncIOMotorCollection
        """
        collection_meta = cls.get_settings().collection_settings
        collection_meta.motor_collection.name = collection_meta.name
        return collection_meta.motor_collection
```

I'm logging the issue with you folks as the error doesn't manifest itself when interacting with an actual mongodb so presumably it isn’t an issue Beanie or the Async engine.

If I'm in the wrong place, please let me know. If you want me to have a dig into the code and PR-it then happy to take a look too - I ask as it might be as long reviewing my PR as making the change and merging it yourselves.

denivyruck:
Isn't this an issue with `mongomock-motor` then? Seems [they're not copying the name](https://github.com/michaelkryukov/mongomock_motor/blob/2a3c9d8187afacb873b8cd3bf258d3731704baf7/mongomock_motor/__init__.py#L44)

cypherlou:
Hey @denivyruck, thanks for the response. I'll pop over and see what the `mongomock-motor` guys say/think. 👍

michaelkryukov:
```py
import asyncio
from mongomock_motor import AsyncMongoMockClient
from motor.motor_asyncio import AsyncIOMotorClient
from bson import DBRef

async def test__aggregate_lookup_dbref():
[...]

PR #884, opened 2024-03-20
Add support for DBRef's `$id` filtering

martincpt:
Related to: #759 

Also related to: [michaelkryukov/mongomock_motor#39](https://github.com/michaelkryukov/mongomock_motor/issues/39)

### DIFF
--- a/mongomock/filtering.py
+++ b/mongomock/filtering.py
@@ -214,6 +214,9 @@ def iter_key_candidates(key, doc):
 
     if doc is None:
         return ()
+    
+    if isinstance(doc, DBRef) and key == '$id':
+        return [doc.id]
 
     if isinstance(doc, list):
         return _iter_key_candidates_sublist(key, doc)


mongomock/filtering.py
```diff
@@ -214,6 +214,9 @@ def iter_key_candidates(key, doc):
 
     if doc is None:
         return ()
+    
+    if isinstance(doc, DBRef) and key == '$id':
```
michaelkryukov, line 218: I can suggest to make this check `DBRef and isinstance(doc, DBRef)` and instead of using `doc.id` directly use  more general `return [doc.as_doc().get(key, NOTHING)]`. This will align well with same code from #878

michaelkryukov:
I would also suggest add at least one test for added functionality.

mongomock/filtering.py
martincpt: Can the `DBRef` class itself be a `None`, or why check against it?

I like the more general return though, works perfectly on my tests.

martincpt:
> I would also suggest add at least one test for added functionality.

Unfortunately, I wasn't able to run the tox command on this project and since I couldn't figure out - and didn't found any detailed description - I didn't bother to write any tests.

PR #885, opened 2024-03-21
Allow custom ClientClass argument on `patch`

Caiofcas:
Hi, this came up as a solution to deal with limitations with the `mongomock` api - namely not supporting transactions and sessions. I understand the decision to raise an error from the lib side, but I still think it should be possible to use the other features if the user chooses to. In my case I created a class that "pretends" to initiate a transaction for testing:

```python
class MongoMockClientWithSession(mongomock.MongoClient):
    """Mongomock does not implement sessions, so we mock it here. This means that
    the atomicity is not actually tested.
    """

    @contextmanager
    def start_session(self, *args, **kwargs):
        yield self

    @contextmanager
    def start_transaction(self, *args, **kwargs):
        yield
```

And used it in my tests with `mongomock.patch`:

```python
@pytest.fixture(autouse=True)
def mock_mongo(): 
    with patch_mongoclient("testhost", "create", ClientClass=MongoMockClientWithSession):
            yield
```

So the transactions are not actually tested but can still perform the tests on the other code functionality.

If this is not wanted no problem, just thought it could be a nice feature.

### DIFF
--- a/mongomock/__init__.pyi
+++ b/mongomock/__init__.pyi
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Literal, Sequence, Tuple, Union
+from typing import Any, Callable, Literal, Sequence, Tuple, Union, TypeVar, Type
 from unittest import mock
 
 from bson.objectid import ObjectId as ObjectId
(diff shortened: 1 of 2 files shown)


Caiofcas:
The code was not right. Sorry for the spam.

PR #894, opened 2024-07-22
Remove Deprecation Warning for Datetime

JohnGrubba:


mongomock/helpers.py
```diff
@@ -1,6 +1,6 @@
 from collections import abc
 from collections import OrderedDict
-from datetime import datetime, timedelta, tzinfo
```
mdomke, line 3: The `UTC` alias has only been added in Python 3.11, but we still have to support Python 3.8. We can use `timezone.utc` instead.

https://docs.python.org/3/library/datetime.html#datetime.UTC

mdomke (changes requested):

PR #895, opened 2024-08-02
fix: match signature of pymongo's collation

minemidnight:
change collation to match the signature of https://github.com/mongodb/mongo-python-driver/blob/3e5387e0ac77131c00b0912367e2237225d18d76/pymongo/synchronous/cursor.py#L865

### DIFF
--- a/mongomock/collection.py
+++ b/mongomock/collection.py
@@ -7,6 +7,7 @@ import functools
 import itertools
 import json
 import math
+from typing import Any
 from packaging import version
 import time
 import warnings
@@ -25,8 +26,11 @@ try:
     from pymongo.operations import IndexModel
     from pymongo import ReadPreference
     from pymongo import ReturnDocument
+    from pymongo.collation import Collation, validate_collation_or_none
+
     _READ_PREFERENCE_PRIMARY = ReadPreference.PRIMARY
 except ImportError:
+
     class IndexModel(object):
         pass
 
@@ -35,6 +39,15 @@ except ImportError:
         AFTER = True
 
     from mongomock.read_preferences import PRIMARY as _READ_PREFERENCE_PRIMARY
+
+    class Collation(object):
+        pass
+
+    def validate_collation_or_none(
+        value: Mapping[str, Any] | Collation | None
+    ) -> dict[str, Any] | None:
+        return value
+
 
 from sentinels import NOTHING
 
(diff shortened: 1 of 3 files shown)


mongomock/collection.py
```diff
@@ -36,6 +38,14 @@ class ReturnDocument(object):
 
     from mongomock.read_preferences import PRIMARY as _READ_PREFERENCE_PRIMARY
 
+    class Collation(object):
+        pass
+
+    def validate_collation_or_none(
+        value: Mapping[str, Any] | Collation | None
```
mdomke: We are still supporting Python versions below 3.10 so that the union operator can not be used but rather `typing.Union` has to be employed.

mongomock/collection.py
```diff
@@ -2024,9 +2034,17 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def alive(self):
         return self._emitted != len(self._compute_results(with_limit_and_skip=False))
 
-    @property
-    def collation(self):
-        return self._collation
+    def collation(self, collation: Collation | Mapping[str, Any]) -> "Cursor":
+        """Adds a :class:`~pymongo.collation.Collation` to this query.
+        Raises :exc:`TypeError` if `collation` is not an instance of
```
mdomke, line 2040: The summary line of a multi-line docstring has to be separated with a blank line

https://peps.python.org/pep-0257/#multi-line-docstrings

mdomke (changes requested):

mdomke:
I know, it's super annoying but the [Type Hinting for Generics in Standard Collections](https://peps.python.org/pep-0585/) has only been introduced in Python 3.9 and since we have to support Python 3.8 for a month or so. Until then, you would have to use `typing.Mapping` in the type annotations you added. That's why you're getting the
```
TypeError: 'ABCMeta' object is not subscriptable
```
error.

mbyrne00:
Hey @mdomke - given it's been over 2 months since your last comment to wait "a month or so", are we ready to proceed with type hints or does this lib still need to continue supporting python 3.8?

mdomke:
@mbyrne00 Python 3.8 has transitioned into EOL-state at the end of October 2024 and I removed support for it recently in #908. I somehow didn't see that @minemidnight adapted the pull-request in the meantime to use `Union` and `Mapping` for backwards-compatible type annotations. Nevertheless, I would like the pipeline to succeed, before I merge the pull-request. In #909 I reformatted the codebase with `ruff`, so that the easiest approach would be to merge `develop` into this topic-branch and run `hatch fmt`.

PR #896, opened 2024-08-02
Add support for $sortyCount aggregation command

userlerueda:
- draft for support for $sortByCount in the $facet stage

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
       context: .
     image: mongomock:latest
     working_dir: /project
-    command: sleep infinity
+    command: hatch test
     environment:
       - TEST_MONGO_HOST=mongodb://mongo/mock
       - MONGODB=5.0.5
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
       context: .
     image: mongomock:latest
     working_dir: /project
-    command: hatch test
+    command: sleep infinity
     environment:
       - TEST_MONGO_HOST=mongodb://mongo/mock
       - MONGODB=5.0.5
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -1364,28 +1364,15 @@ def _handle_sample_stage(in_collection, unused_database, options):
 
 
 def _handle_sort_by_count_stage(in_collection, unused_database, options: Union[str, Dict]):
-    key_count = {}
-    sort_by_count = []
-
     if isinstance(options, Dict) and '$mergeObjects' in options:
         raise NotImplementedError(
             "Although '$mergeObjects' is a valid operator for the $sortByCount stage, "
             'it is currently not implemented in Mongomock.'
         )
-    field_to_count = options[1:]
+    field_to_count = options.lstrip("$")
 
-    for doc in in_collection:
-        field_value = doc.get(field_to_count)
-        if field_value is None:
-            continue
-        if field_value not in key_count:
-            key_count[field_value] = 0
-        key_count[field_value] += 1
-    
-    for value, count in key_count.items():
-        sort_by_count.append({'_id': value, 'count': count})
-    sort_by_count = sorted(sort_by_count, key=lambda x: x['count'], reverse=True)
-    return sort_by_count
+    counter = collections.Counter([doc[field_to_count] for doc in in_collection if field_to_count in doc])
+    return [{"_id": key, "count": count} for key, count in counter.most_common()]
 
 
 def _handle_sort_stage(in_collection, unused_database, options):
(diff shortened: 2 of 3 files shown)


userlerueda:
@mdomke this is my first contribution to the project, I wonder if there is a utility that ignores the order for things such as $sort, in this case the order if relevant for the count, but in the case where count is equal, the order becomes irrelevant, I don't know if there is some helper function that considers this.

Please do let me know, I am happy to add anything that you see necessary help the feature become supported.

mongomock/aggregate.py
```diff
@@ -1364,6 +1365,31 @@ def _handle_sample_stage(in_collection, unused_database, options):
...
+        )
+    field_to_count = options[1:]
+
+    for doc in in_collection:
+        field_value = doc.get(field_to_count)
+        if field_value is None:
+            continue
+        if field_value not in key_count:
+            key_count[field_value] = 0
+        key_count[field_value] += 1
```
mdomke: Hmm... This is probably what [collections.Counter](https://docs.python.org/3/library/collections.html#collections.Counter) is good for. I think you could rewrite all of this as
```python
counter = Counter([doc[group_by] for doc in in_collection if group_by in doc])
return [{"_id": key, "count": count for key, count in counter.most_comon()]
```

mongomock/aggregate.py
```diff
@@ -1364,6 +1365,31 @@ def _handle_sample_stage(in_collection, unused_database, options):
...
+def _handle_sort_by_count_stage(in_collection, unused_database, options: Union[str, Dict]):
+    key_count = {}
+    sort_by_count = []
+
+    if isinstance(options, Dict) and '$mergeObjects' in options:
+        raise NotImplementedError(
+            "Although '$mergeObjects' is a valid operator for the $sortByCount stage, "
+            'it is currently not implemented in Mongomock.'
+        )
+    field_to_count = options[1:]
```
mdomke: It does the same thing, but it might be explaining a little better what is going on
```python
field_to_count = options.lstrip("$")
```

mongomock/aggregate.py
```diff
@@ -1364,6 +1365,31 @@ def _handle_sample_stage(in_collection, unused_database, options):
     return shuffled[:size]
 
 
+def _handle_sort_by_count_stage(in_collection, unused_database, options: Union[str, Dict]):
+    key_count = {}
+    sort_by_count = []
+
+    if isinstance(options, Dict) and '$mergeObjects' in options:
```
mdomke: I think you have to guard for all situations where `options` is a `dict`, because later down in the code it is assumed that `options` is actually a string. So if `options` was something like `{"foo": "bar"}`, you would get an exception when trying to extract `field_to_count`

mongomock/aggregate.py
```diff
@@ -1581,6 +1607,9 @@ def _handle_out_stage(in_collection, database, options):
 
 
 def _handle_count_stage(in_collection, database, options):
+    print(f"{in_collection=}")
+    print(f"{database=}")
+    print(f"{options=}")
```
mdomke: I guess this was for debugging purposes?

mdomke (changes requested):

mongomock/aggregate.py
userlerueda: Sure. Updating this.

mongomock/aggregate.py
userlerueda: It was, I should have removed this prior to submitting the PR. I was trying to understand how the library works. Removing this.

userlerueda:
I have updated and tested but haven't been able to get rid of the assertion error from the diff that states that it is in the incorrect order.

```
E               AssertionError: Unexpected Diff:
E                       fake.0.pipeline_b.0._id != real.0.pipeline_b.0._id (1 != 2)
E                       fake.0.pipeline_b.1._id != real.0.pipeline_b.1._id (2 != 4)
E                       fake.0.pipeline_b.2._id != real.0.pipeline_b.2._id (4 != 1)
```

Here is what the return looks like:
```
[{'_id': 1, 'count': 2}, {'_id': 2, 'count': 2}, {'_id': 4, 'count': 2}, {'_id': 3, 'count': 1}]
```

Since all of them have a count of 2, technically speaking (or at least up to my knowledge) they could come back from mongodb in any order.

I am not sure if there is a way around this. @mdomke How would you approach it?

mongomock/aggregate.py
userlerueda: Done, I have changed the check to just check for dictionaries, and have changed the message to let the user know that using a dictionary for a value in $sortByCount is not yet supported.

docker-compose.yml
```diff
@@ -4,7 +4,7 @@ services:
       context: .
     image: mongomock:latest
     working_dir: /project
-    command: hatch test
+    command: sleep infinity
```
mdomke: I suppose this was needed for manual testing?

mdomke (changes requested):

mdomke:
Can you maybe rebase your changes on the current development version, so that we actually have a test-pipeline where I could look at the error? The tip of the `develop`-branch uses GitHub Actions.

userlerueda:
Sure, let me do that.

docker-compose.yml
userlerueda: Done.

userlerueda:
@mdomke I just rebased against the latest `develop` branch. Please let me know if there is anything else I should do.

PR #920, opened 2025-05-19
add examples basic query, and aggregation with mongomock

federicsp:
This PR adds runnable example scripts inside the examples/ directory to demonstrate usage of mongomock:

The examples can be run with python -m examples.<module_name>, providing an easy way for users to understand how to test MongoDB interactions without a real database.

Additionally, the README is updated.


engFelipeMonteiro:
I am not a developer here, but I think its better if you write the code as a test using unittest syntax and invoke using pytest

PR #838, opened 2023-03-26
Add support to $type operator

JPDevelop:
Added support to mongo's `$type` operator for:
- int
- long
- bool
- string
- object
- null
- missing

PR #929, opened 2025-07-06
Add support for $type aggregation operator

YotamAlon:
Hey All,

Following up on https://github.com/mongomock/mongomock/pull/838 - I wanted this feature as well and since I saw  @JPDevelop didn't respond I thought I'd make the requested change.

Please let me know if any changes are needed.

-original text-

Added support to mongo's $type operator for:

    int
    long
    bool
    string
    object
    null
    missing
    date

### DIFF
--- a/tests/test__mongomock.py
+++ b/tests/test__mongomock.py
@@ -3676,6 +3676,41 @@ class MongoClientAggregateTest(_CollectionComparisonTest):
             ]
         )
 
+    def test__aggregate_type(self):
+        self.cmp.do.insert_one(
+            {
+                '_id': 1,
+                'list': [1, 2, 3],
+                'tuple': (1, 2, 3),
+                'empty_list': [],
+                'empty_tuple': (),
+                'date': datetime.datetime(1999, 12, 19, 1, 2, 3),
+                'int': 3,
+                'str': '123',
+                'bool': True,
+                'none': None,
+            }
+        )
+        self.cmp.compare.aggregate(
+            [
+                {
+                    '$project': {
+                        '_id': False,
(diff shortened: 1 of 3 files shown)


YotamAlon:
I see there are a couple of failing tests, they do not seem to be related to the change here, but I'll see if I can fix them.

YotamAlon:
Fixed the tests!
I did not verify how the issue was introduced, but luckily they all happened in pymongo comparison tests, so it was easy to know what was the right behavior.

grinuri:
@YotamAlon Great fix! Any estimation when it will be merged&delivered?

YotamAlon:
@grinuri I wish I knew. It's up to the new maintainers, whoever they are

tests/test__mongomock.py
```diff
@@ -3676,6 +3673,41 @@ def test__aggregate_is_array(self):
...
+            }
+        )
+        self.cmp.compare.aggregate(
+            [
+                {
+                    '$project': {
+                        '_id': False,
+                        'list': {'$type': '$list'},
+                        'tuple': {'$type': '$tuple'},
+                        'string': {'$type': '$string'},
```
grinuri: @YotamAlon how this test passes ? there is a mismatch between `string` and `str`.

grinuri: @YotamAlon up

YotamAlon: Sorry about the delay, I think it passes because both mongo and mongomock returned the same for the non-existing field `string`.

Anyway, don't know what I was thinking when I wrote this test, but it's fixed now I hope.

PR #950, opened 2026-06-18
Support adding millisecond offsets to dates

### LINKED ISSUE #688: Adding time interval to date objects in aggregation pipeline not working
I want to be able to add some time interval (hours in my case) given by one field _hour_, to a date given by another field _(date)_ to create a new field (_datetime_) in an aggregation pipeline , i.e. performing the equivalent operation to `datetime.datetime(year, month, day) + datetime.timedelta(hours=hour)`. From what I know there is no such timedelta object in bson **(please correct me if I'm wrong and there is a better way to achieve that goal**), but it is possible in mongodb/pymongo to convert the desired time interval into milliseconds and adding this to a date/datetime object. My code looks as follows:
```python
from datetime import datetime

import mongomock
import pymongo

collection = pymongo.MongoClient().db.test
collection.insert_one({
    "date": datetime(2021, 1, 1),
    "hour": 6
})

collection.aggregate([
    {
        "$set": {
            "datetime": {
                "$add": [
                    "$date",
                    {"$multiply": ["$hour", 3600000]} # convert hours to milliseconds
                ]
            }
        }
    }
])
```
The code is working as expected since pymongo "knows" how to add an integer (milliseconds) to a datetime object. However when I want to test my code, using the mocked version of the database, it is not working. Replacing the 6th line of my code to mock the database `collection = mongomock.MongoClient().db.test` results in an TypeError caused by mongomock trying to add an integer to a datetime object.
I'm getting the following Traceback:
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-4e9537e5d845> in <module>
      9 })
     10 
---> 11 collection.aggregate([
     12     {
     13         "$set": {

~/venv/lib/python3.8/site-packages/mongomock/collection.py in aggregate(self, pipeline, session, **unused_kwargs)
   1746     def aggregate(self, pipeline, session=None, **unused_kwargs):
   1747         in_collection = [doc for doc in self.find()]
-> 1748         return aggregate.process_pipeline(in_collection, self.database, pipeline, session)
   1749 
   1750     def with_options(

~/venv/lib/python3.8/site-packages/mongomock/aggregate.py in process_pipeline(collection, database, pipeline, session)
   1401                     "Although '%s' is a valid operator for the aggregation pipeline, it is "
   1402                     'currently not implemented in Mongomock.' % operator)
-> 1403             collection = handler(collection, database, options)
   1404 
   1405     return command_cursor.CommandCursor(collection)

~/venv/lib/python3.8/site-packages/mongomock/aggregate.py in _handle_add_fields_stage(in_collection, unused_database, options)
   1301         for in_doc, out_doc in zip(in_collection, out_collection):
   1302             try:
-> 1303                 out_value = _parse_expression(value, in_doc, ignore_missing_keys=True)
   1304             except KeyError:
   1305                 continue

~/venv/lib/python3.8/site-packages/mongomock/aggregate.py in _parse_expression(expression, doc_dict, ignore_missing_keys)
    830             if it is possible.
    831     """
--> 832     return _Parser(doc_dict, ignore_missing_keys=ignore_missing_keys).parse(expression)
    833 
    834 

~/venv/lib/python3.8/site-packages/mongomock/aggregate.py in parse(self, expression)
    211         for k, v in six.iteritems(expression):
    212             if k in arithmetic_operators:
--> 213                 return self._handle_arithmetic_operator(k, v)
    214             if k in project_operators:
    215                 return self._handle_project_operator(k, v)

~/venv/lib/python3.8/site-packages/mongomock/aggregate.py in _handle_arithmetic_operator(self, operator, values)
    293             return abs(self.parse(values))
    294         if operator == '$add':
--> 295             return sum(self.parse(value) for value in values)
    296         if operator == '$ceil':
    297             return math.ceil(self.parse(values))

TypeError: unsupported operand type(s) for +: 'int' and 'datetime.datetime'
```
I think a type check in the last if statement from the traceback `if operator == '$add'` and probably on other places would help to mirror the behavior of pymongo when dealing with arithmetics of datetime objects and numbers.

gaoflow:
Fixes #688.

MongoDB allows `$add` to combine one date with numeric operands, treating the numeric values as millisecond offsets. Mongomock currently requires every `$add` operand to be numeric, so the issue reproduction raises an assertion instead of producing the shifted datetime.

This keeps the existing numeric-only `$add` behavior, and adds support for a single datetime plus numeric millisecond offsets. The regression test covers the issue's `$multiply` expression and the same operands in reversed order.

Verification:
- `.venv/bin/hatch fmt --check`
- `.venv/bin/python -m pytest tests/test__collection_api.py`
- `.venv/bin/hatch test tests/test__collection_api.py::CollectionAPITest::test__aggregate_add_date_and_milliseconds`
- Manual reproduction from #688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Removed
 - Remove support for deprecated Python version 3.8
 
+### Fixed
+- Support adding numeric millisecond offsets to datetimes in aggregation `$add` expressions.
+
 
 ## [4.3.0] - 2024-11-16
 ### Added
--- a/mongomock/aggregate.py
+++ b/mongomock/aggregate.py
@@ -411,12 +411,25 @@ class _Parser:
 
         parsed_values = list(self.parse_many(values))
         assert parsed_values, f'{operator} must have at least one parameter'
+        if operator == '$add':
+            date_value = None
+            sum_value = 0
+            for value in parsed_values:
+                if value is None:
+                    return None
+                if isinstance(value, datetime.datetime):
+                    assert date_value is None, f'{operator} only accepts one date'
+                    date_value = value
+                else:
+                    assert isinstance(value, numbers.Number), f'{operator} only uses numbers'
+                    sum_value += value
+            if date_value is not None:
+                return date_value + datetime.timedelta(milliseconds=sum_value)
+            return sum_value
         for value in parsed_values:
             if value is None:
                 return None
             assert isinstance(value, numbers.Number), f'{operator} only uses numbers'
-        if operator == '$add':
-            return sum(parsed_values)
         if operator == '$multiply':
             return functools.reduce(lambda x, y: x * y, parsed_values)
 
--- a/tests/test__collection_api.py
+++ b/tests/test__collection_api.py
@@ -4487,6 +4487,25 @@ class CollectionAPITest(TestCase):
                 )
             )
 
+    def test__aggregate_add_date_and_milliseconds(self):
+        self.db.collection.insert_one({'date': datetime(2021, 1, 1), 'hour': 6})
+        actual = self.db.collection.aggregate(
+            [
+                {
+                    '$project': {
+                        '_id': False,
+                        'datetime': {'$add': ['$date', {'$multiply': ['$hour', 3600000]}]},
+                        'datetime_reversed': {'$add': [{'$multiply': ['$hour', 3600000]}, '$date']},
+                    }
+                }
+            ]
+        )
+        expected = {
+            'datetime': datetime(2021, 1, 1, 6),
+            'datetime_reversed': datetime(2021, 1, 1, 6),
+        }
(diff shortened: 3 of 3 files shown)


binggao1230:
CI triage note: the current red jobs look unrelated to this `$add` date change.

- `test (3.9)` fails before running tests while resolving `coverage[toml]>=7.11`, because the available coverage versions in that range require Python >=3.10.
- The 3.10-3.13 jobs are cancelled after the matrix failure, but their logs show the same three existing regex comparison failures (`test__find_by_regex_object`, `test__find_by_regex_string`, `test__aggregate_regexpmatch`). I found the same regex failures in an unrelated earlier run: https://github.com/mongomock/mongomock/actions/runs/25632285652/job/75238039178

The checks directly covering this change pass locally:

- `hatch fmt --check`
- `python -m pytest tests/test__collection_api.py`
- `hatch test tests/test__collection_api.py::CollectionAPITest::test__aggregate_add_date_and_milliseconds`

